PR #7, opened 2022-05-27
mobile friendly mods

LateBy5:
Updated the memory component and cpu page to be slightly more mobile friendly. Tested various viewports out locally - changes look alright.

### DIFF
--- a/src/components/screen.ts
+++ b/src/components/screen.ts
@@ -55,8 +55,7 @@ export const Screen = FC<
     figure(
       {
         style: {
-          width: "100%",
-          marginInline: "auto",
+          textAlign: "center",
           borderTop: "2px solid gray",
           borderLeft: "2px solid gray",
           borderBottom: "2px solid lightgray",
--- a/src/components/screen.ts
+++ b/src/components/screen.ts
@@ -34,7 +34,6 @@ export const Screen = FC<
 >("hack-screen", (el, { memory }) => {
   const screen = (el[State].screen ??= canvas({ width: 512, height: 256 }));
   const ctx = (el[State].ctx ??= screen.getContext("2d") ?? undefined);
-  el.style.width = "518px";
 
   if (ctx) {
     const image = assertExists(
@@ -56,11 +55,12 @@ export const Screen = FC<
     figure(
       {
         style: {
+          width: "100%",
+          marginInline: "auto",
           borderTop: "2px solid gray",
           borderLeft: "2px solid gray",
           borderBottom: "2px solid lightgray",
-          borderRight: "2px solid lightgray",
-          marginBottom: "0",
+          borderRight: "2px solid lightgray"
         },
       },
       screen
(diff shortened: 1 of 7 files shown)


src/pages/cpu.ts
```diff
@@ -69,23 +69,27 @@ export const CPU = (
...
-            justifyContent: "center",
+            gridTemplateColumns: "repeat(4, minmax(125px, 283px))",
+            justifyContent: "center"
           },
           "hack-screen": {
-            gridColumn: "3 / span 2",
-          },
+            width: "100%",
+            maxWidth: "600px",
+            gridColumn: "3 / span 4",
```
DavidSouther, line 78: It looks like this can be 3 / span 2

src/pages/cpu.ts
```diff
@@ -69,23 +69,27 @@ export const CPU = (
...
+          }
         },
         "@media (max-width: 1195px)": {
```
DavidSouther, line 81: I know you found it this way, but this is backwards - we want to only use the min-width media selectors, inline with the breakpoints in picocss.

src/pages/cpu.ts
```diff
@@ -69,23 +69,27 @@ export const CPU = (
...
           "div.grid": {
-            gridTemplateColumns: "repeat(4, 283px)",
-            justifyContent: "center",
+            gridTemplateColumns: "repeat(4, minmax(125px, 283px))",
+            justifyContent: "center"
           },
           "hack-screen": {
-            gridColumn: "3 / span 2",
-          },
+            width: "100%",
```
DavidSouther, line 76: This width turns out we don't need

src/components/screen.ts
```diff
@@ -56,11 +55,12 @@ export const Screen = FC<
     figure(
       {
         style: {
+          width: "100%",
+          marginInline: "auto",
```
DavidSouther, line 59: marginInline is the secret sauce, here, and we can remove the width to keep the border tight around the canvas

DavidSouther (changes requested):

src/pages/cpu.ts
LateBy5: Our version of pico css doesn't seem to contain media queries for the container class, so I shimmed them in the js for the time being

src/components/screen.ts
LateBy5: > marginInline is the secret sauce, here, and we can remove the width to keep the border tight around the canvas

Turns out we don't need either. Instead I'm using text-align to center the canvas within the figure element - that way nothing gets screwy with the horizontal scrolling

LateBy5:
Per your initial comments, I took a first look at pico css (lol - should of did that from the start). I updated the styling to use / align with the ethos of pico css

src/pages/cpu.ts
DavidSouther: We could add containers using the query params to pico, but this is I think better without them.

src/components/screen.ts
DavidSouther: Great - and then, the last piece is the border stays visible but doesn't get wider than the canvas.

DavidSouther (approved):

Issue #72, opened 2022-07-13
Clock / Sequential chips


Shimon-Schocken:
I suggest that we implement the clock exactly the same way it is implemented in the desktop version. The new IDE implementation gives the impression that the clock is part of the chip. Actually, the clock is a global "agent" that effects all the sequential chips in the computer.

Definition:
- DFF is a sequential chip (because it relies on the clock)
- A chip that lists a DFF as a chip-part is a sequential chip
- A chip that lists one or more sequential chip-parts is a sequential chip

Note: Nothing in the HDL of a chip mentions that it is sequential. The IDE should figure it out implicitly when it builds the chip's tree, i.e. when the chip is "loaded" into the IDE.

For now, we can put the clock button next to the Eval / Save buttons, and worry about exact layout later.

Here are the Clock button states: 

- The clock button displays the clock value: 0, 0+, 1, 1+, 2, 2+, ...
- The clock button is displayed, always. The default display is dimmed.
- If the loaded chip is sequential, the clock button becomes active (not dimmed)
- If the shown clock value is n, clicking it makes the clock display n+
- If the shown clock value is n+, clicking it makes the clock display n+1

Shimon-Schocken:
Thx, perfect -- Shimon

Issue #28, opened 2022-05-31
Allow editing wide input pins in HDL


DavidSouther:

DavidSouther:
From [this comment](https://github.com/DavidSouther/computron5k/issues/65#issuecomment-1185157905),

>  interacting with binary input values is a major part of the student experience and the overall IDE UX. For example, presently, when you click on a 16-bit binary input, the displayed 16-bit value increments by 1. [...] When the user clicks a bit in any n-bit input pint, the bit should flip (and only that bit).

PR #73, opened 2022-07-16
Feat/busedit

DavidSouther:
As per https://github.com/DavidSouther/computron5k/issues/28 and https://github.com/DavidSouther/computron5k/issues/72

image


DavidSouther:
Close #28 
Close #72

PR #191, opened 2023-08-27
ALU visualization

AntoinePiret:
New SVG visualization for the ALU components
Update of the operation symbols to fit to the book (x, y, x+y, etc...)


AntoinePiret:
Changed the color of the internal pins to gray (Shimon's request)

PR #201, opened 2023-12-16
CPU page improvements + Assembler page

netalondon:


components/src/runbar.tsx
```diff
@@ -1,15 +1,47 @@
...
 import { Timer } from "@nand2tetris/simulator/timer.js";
+import { ChangeEvent, ReactNode, useEffect, useState } from "react";
 import { useTimer } from "./timer.js";
 
-export const Runbar = (props: { runner: Timer; children?: ReactNode }) => {
+export const Runbar = (props: {
+  runner: Timer;
+  prefix?: ReactNode;
+  children?: ReactNode;
+  overrideTooltips?: (string | undefined)[];
```
DavidSouther, line 9: Please expand this type out to the known tooltips - 

```
interface RunbarTooltipOverrides {
   play: string;
}
```

and then in the component declaration

```
const Runbar = (props: {overrideTooltips?: Partial<RunbarTooltipOverrides>}) {}
```

components/src/stores/asm.store.ts
```diff
@@ -0,0 +1,343 @@
...
+import { Dispatch, MutableRefObject, useContext, useMemo, useRef } from "react";
+import { useImmerReducer } from "../react.js";
+import { BaseContext } from "./base.context.js";
+
+function defaultSymbols(): [string, string][] {
+  return [
+    ["R0", "0"],
+    ["R1", "1"],
+    ["R2", "2"],
+    ["...", ""],
```
DavidSouther, line 24: Should this be the rest of `R3` to `R15`?

components/src/stores/asm.store.ts
```diff
@@ -0,0 +1,343 @@
...
+  resultHighlight: Span | undefined;
+  sourceHighlight: Span | undefined;
+  highlightMap: Map<Span, Span>;
+}
+
+class Translator {
+  asm: Asm | null = null;
+  current = -1;
+  done = false;
+  symbols: [string, string][] = [];
```
DavidSouther, line 41: Prefer explicit types to arrays-as-tuples:

```
interface TranslatorSymbol {
  name: string;
  value: string;
}
```

components/src/stores/asm.store.ts
```diff
@@ -0,0 +1,343 @@
...
+  hiddenSymbols: Map<number, [string, boolean]> = new Map();
```
DavidSouther, line 42: Again, prefer explicit types - I _think_ `[string, boolean]` here could also be `Set<string>`?

components/src/stores/asm.store.ts
```diff
@@ -0,0 +1,343 @@
...
+    this.asm = asm;
+    fillLabel(asm, (name, value, isVar) => {
+      if (isVar) {
+        this.hiddenSymbols.set(value, [name, false]);
+      } else {
+        this.symbols.push([name, value.toString()]);
+      }
+    });
+    asm.instructions = asm.instructions.filter(
+      (instruction) => instruction.type !== "L"
```
DavidSouther, line 58: (Optional) For fun, this can be a bit shorter with (IMO) no loss of readability:

```
asm.instructions = asm.instructions.filter(({type}) => type !== 'L');
```

components/src/stores/asm.store.ts
```diff
@@ -0,0 +1,343 @@
...
+}
+
+interface HighlightInfo {
+  resultHighlight: Span | undefined;
+  sourceHighlight: Span | undefined;
+  highlightMap: Map<Span, Span>;
+}
+
+class Translator {
+  asm: Asm | null = null;
```
DavidSouther, line 38: I would suggest not allowing `null` here, and always having `asm` be valid. I believe `{instructions: []}` is always a valid `ASM`, and makes several checks throughout the class unnecessary.

components/src/stores/asm.store.ts
```diff
@@ -0,0 +1,343 @@
...
+        currentLine++;
+      }
+    }
+  }
+
+  step(highlightInfo: HighlightInfo) {
+    if (this.asm === null || this.current >= this.asm.instructions.length - 1) {
+      return;
+    }
+    this.current++;
```
DavidSouther, line 83: (optional, personal style) This is a weird personal thing for me, but I don't like using `++` outside `for` loop counters because, and again, this is a me thing, I read `++` as  "increment pointer" and we're not dealing with pointers in JS. So I would write this as `this.current += 1`.

components/src/stores/asm.store.ts
```diff
@@ -0,0 +1,343 @@
...
+    const instruction = this.asm.instructions[this.current];
+    if (instruction.type === "A" || instruction.type === "C") {
+      highlightInfo.sourceHighlight = instruction.span;
+      const result = translateInstruction(this.asm.instructions[this.current]);
+      if (result === undefined) {
+        return;
+      }
+      this.result += `${bin(result)}\n`;
```
DavidSouther, line 91: For the sizes of programs we're dealing with, this shouldn't be an issue (and I haven't run benchmarks on this recently), but in my experience it is more performant in JavaScript engines to manage arrays of small strings, and then `.join("\n")` at the end, rather than modifying large strings in place.

simulator/src/languages/asm.ts
```diff
@@ -33,11 +32,21 @@ export type AsmAInstruction = AsmALabelInstruction | AsmAValueInstruction;
...
 
 export interface AsmAValueInstruction {
   type: "A";
   value: number;
+  span: Span;
+  lineNum: number;
+}
+
+export function isAValueInstruction(
+  inst: AsmAInstruction
```
DavidSouther, line 47: (Optional) We could probably widen this type to AsmInstruction (and in isALabelInstruction) and include the `type == 'A'` check here to make it easier for callers.

simulator/src/languages/grammars/tst.ohm
```diff
@@ -23,7 +23,7 @@ Tst <: Base {
   TstSetOperation = Set Name Index? Number
   Index = OpenSquare wholeDec? CloseSquare
   Condition = Value CompareOp Value
-  TstEvalOperation = Eval | Tick | Tock | TickTock
+  TstEvalOperation = Eval | TickTock| Tick | Tock
```
DavidSouther, line 26: Ohhh yes this is critical haha! Please add a space after `TickTock` (and below)

simulator/src/timer.ts
```diff
@@ -11,7 +11,7 @@ export abstract class Timer {
   }
 
   /// Update the simulation state, but DO NOT perform any UI changes.
-  abstract tick(): boolean;
+  abstract tick(): Promise<boolean>;
```
DavidSouther, line 14: Making `tick` async has some relatively significant performance implications (admittedly, only for the very highest run rates like trying to get to simulations in the MHz range). Could you add a comment as to why you changed it to Promise, so that we can keep note of that going forward? (And know what to look out for when someone tries to make it sync again)

simulator/src/util/asm.ts
```diff
@@ -69,28 +69,37 @@ function aop(asm: string): number {
...
 
   return makeC(mode, cop, aop, jop);
 }
 
 export function makeC(
   isM: boolean,
-  op: COMMANDS_OP,
-  assign: ASSIGN_OP = 0,
-  jmp: JUMP_OP = 0
+  op: COMMANDS_OP | undefined,
```
DavidSouther, line 92: These were written with defaults to make it easier to create the "common" C operations, that is, with no Jump and no Assignment. By marking these `| undefined`, you are requiring callers to provide values, while allowing them also to provide non sensible values. With the intent to disallow invalid C instructions, the API would be better as having these required and forcing callers to explicitly say they don't want assign or jump.

"illegal states should be unrepresentable".

web/public/bitmap_editor.html
DavidSouther, line 1: Do we have permission to use this file from the original copyright holders? If so, please include a copyright block & license at the top, and we can look at re-implementing this "properly" in React at some point in the future.

web/src/pages/bitmap.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const BitmapEditor = () => {
+  const { toolStates } = useContext(AppContext);
+
+  useEffect(() => {
+    toolStates.setTool("bitmap");
+  }, []);
+
+  return (
+    <div style={{ width: "100%", height: "100%" }}>
+      <iframe
```
DavidSouther, line 13: Ah very nice way to include the bitmap editor in context

web/src/pico/icon.scss
```diff
@@ -23,15 +28,15 @@ ul.icon-list {
       opacity: 0;
     }
 
-    &:hover {
-      span,
-      a {
-        top: var(--raise);
-      }
+    // &:hover {
```
DavidSouther, line 31: Feel free to remove unused code, it's in source control if for whatever reason we needed it back.

DavidSouther (changes requested):
As usual, phenomenal work. A couple review comments around a few things, but yeah, really great code here!

components/src/stores/asm.store.ts
DavidSouther: Sounds good, just add a comment with that reasoning

components/src/stores/asm.store.ts
DavidSouther: Oh yeah way more clear this way!

simulator/src/cpu/alu.ts
```diff
@@ -15,12 +14,34 @@ export type COMMANDS_ASM =
...
   | "D-A"
-  | "D-M"
   | "A-D"
   | "D&A"
-  | "D&M"
-  | "D|A"
-  | "D|M";
+  | "D|A";
+
+const commandASMValues: string[] = [
```
DavidSouther, line 21: It looks like you've added the literal constant for for all of these. To avoid repetition, these can be expressed using an `as const` pattern:

```
const value = ['', 'a', 'b', 'c'] as const;
export type VALUES = typeof value[number];
```

Adding `as const` to the array declaration will change the type from `string[]` to `['', 'a', 'b', 'c']`, that is, the array with literally 4 elements with these strings in order. Then, the `typeof array[number]` is the union of the types within the literal array type at any index. (and yes, this means silly things like `string[][number]` is `string`)


BUT!

That leaves these as arrays, and `.include` below is linear. Yes, this is "linear" in an array of 20 items, but we can do better.

```
const values = new Set(['', 'a', 'b', 'c'] as const);
export type VALUES = typeof values extends Set<infer S> ? S : never; // Extracts `''|'a'|'b"|'v'` from the generic type of Set

values.has('b'); // Good to go
values.has('d'); // An error!? Argument of type '"d"' is not assignable to parameter of type '"" | "a" | "b" | "c"'.
values.has('d' as unknown as VALUES); // No type error, good for the isValue()

function isValue(s: unknown): s is VALUES {
  return values.has(s as VALUES);
}
```

DavidSouther (changes requested):

DavidSouther (approved):

PR #204, opened 2023-12-24
Add a warning dialog before reset action

netalondon:


web/src/shell/settings.tsx
```diff
@@ -28,175 +29,223 @@ export const Settings = () => {
...
-          <a
-            style={{ color: "rgba(0, 0, 0, 0)" }}
-            className="close"
-            href="#root"
-            onClick={(e) => {
-              e.preventDefault();
-              settings.close();
+        <main>
+          <div style={{ margin: "10px" }}>
+            {
```
DavidSouther, line 47: No need for {'string'} here, just the plain text node is fine.

web/src/shell/settings.tsx
```diff
@@ -28,175 +29,223 @@ export const Settings = () => {
...
-            <Trans>Settings</Trans>
-          </p>
-          <a
-            style={{ color: "rgba(0, 0, 0, 0)" }}
-            className="close"
-            href="#root"
-            onClick={(e) => {
-              e.preventDefault();
-              settings.close();
+        <main>
```
DavidSouther, line 45: Generally, the structure for article > main can also have an article > footer and article > header. So in this case, I think you can use more structure to get the styling you want

    <article>
      <header>Confirm Reset</header>
      <main>The "reset files" ... </main>
      <footer classname="flex">
        <button>Confirm</button>
        <button>Cancel</button>
      </footer>
    </article>

DavidSouther (approved):

PR #212, opened 2024-01-02
Fix assembler scroll bug

netalondon:
Bug description: the assembler editor would autoscroll to top every time an edit was made.

### DIFF
--- a/web/src/pages/asm.tsx
+++ b/web/src/pages/asm.tsx
@@ -174,15 +174,13 @@ export const Asm = () => {
               return;
             }
             sourceCursorPos.current = index;
-            actions.updateHighlight(index, true);
+            setTimeout(() => {
+              actions.updateHighlight(index, true);
+            }, 100);
           }}
           grammar={ASM.parser}
           language={"asm"}
-          highlight={
-            state.translating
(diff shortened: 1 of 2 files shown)


web/src/pages/asm.tsx
```diff
@@ -174,15 +174,13 @@ export const Asm = () => {
               return;
             }
             sourceCursorPos.current = index;
-            actions.updateHighlight(index, true);
+            setTimeout(() => {
```
DavidSouther, line 177: Is the intent of the 100ms setTimeout to actually wait 1/10th a second, or is it a "small" timeout for `setTimeout` to make the update happen "after" this render? If the latter, `setImmediate` would make that a bit more obvious. If the former, then please add a comment for why 100ms is the chosen timeout.

DavidSouther (changes requested):

web/src/pages/asm.tsx
DavidSouther: Hmm, interesting comment... so it's not really a timeout, but a mouseUp... the comment is good for now, someone can take a closer look later. :)

PR #213, opened 2024-01-03
Add Fill.asm tests + other small test script fixes

netalondon:


components/src/stores/cpu.store.ts
```diff
@@ -109,9 +113,15 @@ export function makeCpuStore(
...
     },
 
+    setAnimate(value: boolean) {
+      animate = value;
+    },
+
     testStep() {
       const done = test.step();
-      dispatch.current({ action: "testStep" });
+      if (animate || done) {
```
DavidSouther, line 122: While this is more correct, it doesn't feel entirely right? That is, I recover there are more times when we would want to update the state, but at least this does make the current animate behavior work.

web/src/versions.ts
```diff
@@ -1,16 +1,30 @@
...
+  let version = getVersion();
 
   while (version < CURRENT_VERSION) {
-    await versionUpdates[version](fs);
-    version++;
+    try {
+      await versionUpdates[version](fs);
+      version++;
+    } catch (e) {
+      version++;
```
DavidSouther, line 23: This should probably get logged somewhere but we don't have a good logging system

web/src/shell/test_panel.tsx
```diff
@@ -52,7 +54,12 @@ export const TestPanel = ({
     try {
       const path = await filePicker.select();
       const tst = await fs.readFile(path);
-      const cmp = await fs.readFile(path.replace(/\.tst$/, ".cmp"));
+      let cmp: string | undefined = undefined;
+      try {
+        cmp = await fs.readFile(path.replace(/\.tst$/, ".cmp"));
+      } catch (e) {
+        // The doesn't have to be a compare file
```
DavidSouther, line 61: "There"

web/src/versions.ts
```diff
@@ -1,16 +1,30 @@
 import { FileSystem } from "@davidsouther/jiffies/lib/esm/fs";
+import * as project_04 from "@nand2tetris/projects/project_04/index.js";
 
-const CURRENT_VERSION = 1;
+const VERSION_KEY = "version";
+const CURRENT_VERSION = 2;
+
+export function getVersion() {
+  return Number(localStorage.getItem(VERSION_KEY)) ?? 0;
```
DavidSouther, line 8: I think you want the ?? Inside the constrict parens, `Number(local... ?? "0")`

DavidSouther (changes requested):

web/src/versions.ts
DavidSouther: Hmm. So if a version doesn't work, we don't have its content in localstorage, which does put it an an inconsistent (but overall, usually functional) state.  Probably a console.warn and a setStatus.

web/src/shell/settings.tsx
```diff
@@ -1,13 +1,14 @@
-import loaders from "@nand2tetris/projects/loader.js";
 import { i18n } from "@lingui/core";
 import { Trans } from "@lingui/macro";
-import { useContext, useEffect, useMemo } from "react";
 import { BaseContext } from "@nand2tetris/components/stores/base.context.js";
+import loaders from "@nand2tetris/projects/loader.js";
+import { useContext, useEffect, useMemo } from "react";
 import { AppContext } from "../App.context";
 
+import { TrackingDisclosure } from "src/tracking";
```
DavidSouther, line 8: I think these should be `../` instead of `src/`. for consistency

DavidSouther (changes requested):

web/src/versions.ts
DavidSouther: Definitely do not ever use alert. It is highly disruptive to any browser based workflow, and we should all just forget it exists.

Because we have ways to reset the loaded files, I think the best answer today is to use 

```console.warn(`Error loading files at version ${version}\`, e)```

And at some point we'll come back and add proper application analytics logging.

DavidSouther (approved):

PR #216, opened 2024-01-07
Eliminate discrepencies between online IDE and desktop version

netalondon:
This handles cases where the online IDE permits HDL code that is considered an error by the desktop java version.

### DIFF
--- a/components/src/stores/chip.store.ts
+++ b/components/src/stores/chip.store.ts
@@ -321,7 +321,7 @@ export function makeChipStore(
 
     async compileChip(hdl: string) {
       chip.remove();
-      const maybeChip = await parseChip(hdl);
+      const maybeChip = await parseChip(hdl, chipName);
       if (isErr(maybeChip)) {
         const error = Err(maybeChip);
         setStatus(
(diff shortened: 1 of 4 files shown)


components/src/stores/chip.store.ts
```diff
@@ -335,12 +335,7 @@ export function makeChipStore(
         });
         return;
       }
-      if (Ok(maybeChip).name != chipName) {
-        setStatus("Warning: Chip name doesn't match selected chip");
```
DavidSouther, line 339: I understand the top level statement is "eliminate discrepancies", but is that universally what we want? In this case, I chose to deliberately introduce a more specific error message. In general, my desire with the web ide was to be "better" than the Java desktop app, including more descriptive error messages, etc.

simulator/src/chip/builder.ts
```diff
@@ -95,10 +96,103 @@ interface InternalPin {
...
+  if (isConstant(rhs.pin)) {
+    return Err({
+      message: `Cannot write to constant bus`,
+      span: rhs.span,
+    });
+  }
+  return Ok(true);
+}
+
 export async function build(
```
DavidSouther, line 184: This function is now in my estimate at the limit of its complexity. The next time someone touches it, it should get refactored first :)

simulator/src/languages/hdl.ts
```diff
@@ -30,6 +30,7 @@ export interface Part {
 
 export interface HdlParse {
   name: string;
+  nameSpan: Span;
```
DavidSouther, line 33: I think I'd rather see this as `name: {value: string, span?: Span}`

DavidSouther (changes requested):

components/src/stores/chip.store.ts
DavidSouther: Ahh gotcha, ok I'm caught up. Yes, I was explicitly flagging losing the specific error, but I see now that it's in build (which we agree is at the limits of complexity).

DavidSouther (approved):

PR #218, opened 2024-01-12
Start work on VM Builtins

DavidSouther:


DavidSouther:
Not sure why that test is failing

PR #219, opened 2024-01-12
Allow only lowercase boolean literals

netalondon:


simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
+  private isConstant(pinName: string): boolean {
+    return (
+      pinName === "false" ||
+      pinName === "true" ||
+      pinName === "0" ||
+      pinName === "1"
+    );
+  }
+
+  public async build() {
```
DavidSouther, line 130: (nit) public is default, no need to have it in the signature

simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
+    this.chip = new Chip(
+      parts.ins.map(({ pin, width }) => ({ pin: pin.toString(), width })),
+      parts.outs.map(({ pin, width }) => ({ pin: pin.toString(), width })),
+      parts.name.value,
+      [],
+      parts.clocked
+    );
+  }
+
+  private isConstant(pinName: string): boolean {
```
DavidSouther, line 121: No use of `this`, no need to be in the class.

simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
+  private parts: HdlParse;
+  private fs?: FileSystem;
+  private expectedName?: string;
+
+  private chip: Chip;
+  private internalPins: Map<string, InternalPin> = new Map();
+  private inPins: Map<string, Set<number>> = new Map();
+  private outPins: Map<string, Set<number>> = new Map();
+
+  public constructor(parts: HdlParse, fs?: FileSystem, name?: string) {
```
DavidSouther, line 108: (nit) No need for explicit public

simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
+        message: `Wrong chip name`,
+        span: this.parts.name.span,
+      });
+    }
+
+    if (this.parts.parts === "BUILTIN") {
+      return getBuiltinChip(this.parts.name.value);
+    }
+
+    try {
```
DavidSouther, line 142: In this project, I generally prefer returning a Result type to try/catch, though I do admit it hasn't been done consistently. Based on eg the changes to the undefined chip name, I think sticking with the original idea and checking `if !isErr(result) return Ok(this.chip)` would be good?

simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
-  parts: HdlParse,
-  fs?: FileSystem,
-  name?: string
-): Promise<Result<Chip, CompilationError>> {
-  if (name && parts.name.value != name) {
-    return Err({
-      message: `Wrong chip name`,
-      span: parts.name.span,
-    });
+  private createWire(lhs: PinParts, rhs: PinParts): Connection {
```
DavidSouther, line 185: No `this`, can be a bare function

simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
-            if (pinData.isDefined) {
-              return Err({
-                message: buildChip.isOutPin(rhs.pin)
-                  ? `Cannot write to output pin ${rhs.pin} multiple times`
-                  : `Internal pin ${rhs.pin} already defined`,
-                span: rhs.span,
-              });
-            }
-            pinData.isDefined = true;
+  private getIndices(pin: PinParts): number[] {
```
DavidSouther, line 303: No `this`

simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
+      const indices = [];
+      for (let i = pin.start; i <= pin.end; i++) {
+        indices.push(i);
+      }
+      return indices;
+    }
+    return [-1];
+  }
+
+  private checkMultipleAssignments(
```
DavidSouther, line 314: After extracting `getIndices`, no `this`

simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
+    pin: PinParts,
+    assignedIndexes: Map<string, Set<number>>
+  ) {
```
DavidSouther, line 317: I would prefer `: Result<void>` as the return type, and then `return Ok()` at the end.

simulator/src/chip/builder.ts
```diff
@@ -82,260 +82,279 @@ export async function loadChip(
...
+    let errorIndex: number | null = null;
+    const indices = assignedIndexes.get(pin.pin);
+    if (!indices) {
+      assignedIndexes.set(pin.pin, new Set(this.getIndices(pin)));
+    } else {
+      if (indices.has(-1)) {
+        // -1 stands for the whole bus width
```
DavidSouther, line 324: Move or repeat this comment when declaring errorIndex

simulator/src/chip/builder.ts
DavidSouther, line 1: I really like the changes overall, with a few minor concerns:

1. Prefer keeping `Result` as the return type. You should be able to use `Result<void, CompilationError>` and `Ok()` for signalling success with no information.
2. private functions that don't access `this` don't need to be in the class, and should be just bare functions in the file.
3. `public` is the default visibility, so we don't add it. (This should probably be a lint check but I didn't think to add it)

DavidSouther (changes requested):

simulator/src/chip/builder.ts
DavidSouther: Hmm Ok - I made a release of jiffies (v2.1.5) that allows `Ok()` and `Result<void>`. You should be able to get it to work that way now. If it's still an issue, let me know.

https://github.com/jefri/jiffies/compare/v2.1.4...v2.1.5

DavidSouther: :( Dang it - try 2.1.6? I think I have the CI right for jiffies now

DavidSouther (approved):

PR #221, opened 2024-01-30
Add download project button

netalondon:


web/src/pages/chip.tsx
```diff
@@ -129,6 +130,31 @@ export const Chip = () => {
     [actions]
   );
 
+  const downloadRef = useRef<HTMLAnchorElement>(null);
+
+  const download = (blob: Blob) => {
```
DavidSouther, line 135: I think these are really one function (or they're inverted for the function call).

First, return early if there's no ref.
Second, prepare the zip file.
Third, put the blob of the zip file in the download link.

It's a bit crowded but IMHO at this point it's less complicated to do it at once. We can refactor later if necessary.

DavidSouther (approved):

DavidSouther (approved):

PR #223, opened 2024-02-01
Enforce equal bus widths

netalondon:
Chip implementation where both sides of a wire had different widths passed in the online IDE, but it should count as an error.


DavidSouther (changes requested):
Could you add some tests for this?

DavidSouther (approved):

PR #224, opened 2024-02-08
Update software suite files

netalondon:
The main changes here, other than the project files themselves, are to the test description language, in order to make its behavior compatible with the book and Java emulator.

Mainly:
1. Separate the concepts of a simulation step (a series of commands ending with a semicolon) from a compound instruction (while or repeat), allowing them to overlap.

2. Treat outputs like DRegister[] correctly following section B.2.4 which states that the syntax chipName[varName] refers to the chip's internal state.

With these changes, the project 5 test scripts pass as is, without the need to adjust them to the online IDE specifically.


simulator/src/compare.ts
```diff
@@ -30,7 +30,7 @@ export function diff(as: string[], bs: string[]): Diff[] {
   for (let col = 0; col < q; col++) {
     const a = as[col] ?? "";
     const b = bs[col] ?? "";
-    if (a !== b && !a.match(/\*+/)) {
+    if (a.trim() !== b.trim() && !a.match(/\*+/)) {
```
DavidSouther, line 33: Would it make more sense to do the trim when reading the value? Does the diff need the untrimmed strings?

simulator/src/languages/grammars/tst.ohm
```diff
@@ -2,9 +2,12 @@ Tst <: Base {
...
 
-  TstRepeat = Repeat Number? OpenBrace TstStatement+ CloseBrace
-  TstWhile = While Condition OpenBrace TstStatement+ CloseBrace
-  TstStatement = List<TstOperation, ","> (Semi | Bang)
+  TstRepeat = Repeat Number? OpenBrace TstCommand+ CloseBrace
+  TstWhile = While Condition OpenBrace TstCommand+ CloseBrace
+  TstStatement = TstCommand
+
+  TstCommand = TstOperation Separator
+  Separator = (Semi | Bang | ",")
```
DavidSouther, line 10: Prefer literall Comma (and add it to base if it isn't already)

simulator/src/cpu/cpu.ts
```diff
@@ -76,6 +76,10 @@ export function cpuTick(
   const a = bits.am ? inM : A;
   const [ALU, flag] = alu(bits.op, D, a);
 
+  if (bits.d2) {
+    D = ALU;
```
DavidSouther, line 80: My understanding was that the D registers should not update values until the `tock` phase of the clock cycle.

Regardless, I believe you here but could you add a comment citing to the book where it makes this clear?

simulator/src/test/tst.ts
```diff
@@ -10,17 +15,24 @@ export abstract class Test<IS extends TestInstruction = TestInstruction> {
   protected _outputList: Output[] = [];
   protected _log = "";
   fs: FileSystem = new FileSystem();
+  protected setStatus?: (status: string) => void;
+
+  constructor(setStatus?: (status: string) => void) {
```
DavidSouther, line 20: The `simulator` should not know about `status`, but `echo` and `clearEcho` kind of mean it has to... I'd like this to be more separate in the future, but maybe it's find for now? Eh, rename it to `doEcho` and it's fine given the Tst interface.

simulator/src/cpu/cpu.ts
DavidSouther: Ok... Let's summarize this in a comment. Something like:

> While a DRegister would update during the Tock clock step, this implementation updates the D internal state during tick because the test will need to access the internal D state.

DavidSouther (approved):

PR #227, opened 2024-02-11
Fix check for wrapping with sub-bus

### LINKED ISSUE #225: [bug]: Order of test set operations affects result when using subbus on input pin
### Program

Hardware Simulator

### Interface

### Contact Details

_No response_

### What happened?

Hi. The following snippet produces errors purely dependent on the order in which inputs are set in a test statement.

```
CHIP ALU {
    IN  
        x[16],
        zx, // zero the x input?
        nx; // negate the x input?
    OUT 
        out[16]; // 16-bit output

    PARTS:
    Not16(in=x, out=notx);
    
    // select x
    // zx nx  out
    // 0  0   x
    // 0  1   !x
    // 1  0   0
    // 1  1   !0 = 1111111111111111 
    Mux4Way16(a=x, b=notx, c[0..15]=false, d[0..15]=true, sel[0]=nx, sel[1]=zx, out=out);
}
```

```
output-list x%B1.16.1 zx%B1.1.1 zx%B1.1.1 out%B1.16.1;

set x %B000000000010001;  // x = 17

// Compute x
set zx 0, set nx 0, eval, output; // out = 17
set nx 0, set zx 0, eval, output; // out = 17 

// Compute !x
set zx 0, set nx 1, eval, output;  // out = -18
set nx 1, set zx 0, eval, output;  // out = -18

// Compute 0
set nx 0, set zx 1, eval, output;  // out = 0
set zx 1, set nx 0, eval, output;  // out = 17 (wrong! - behaving as if zx=0)

// Compute -1
set nx 1, set zx 1, eval, output;  // out = -1
set zx 1, set nx 1, eval, output;  // out = -18 (wrong! - behaving as if zx=0)
```

This anomaly does not occur in the Java version.

### Additional Comments

_No response_

### Code of Conduct

netalondon:
Fixes #225

### DIFF
--- a/simulator/src/chip/chip.tsx
+++ b/simulator/src/chip/chip.tsx
@@ -414,7 +414,7 @@ export class Chip {
     }
 
     // Wrap the chipPin in an OutBus when the part side is dimensioned
-    if (to.start > 0 || to.width !== chipPin.width) {
+    if (to.start > 0 || to.width !== partPin.width) {
       chipPin = new OutSubBus(chipPin, to.start, to.width);
     }
 
@@ -433,7 +433,7 @@ export class Chip {
     from.width ??= chipPin.width;
 
     // Wrap the partPin in an InBus when the part side is dimensioned
-    if (to.start > 0 || to.width !== chipPin.width) {
+    if (to.start > 0 || to.width !== partPin.width) {
       partPin = new InSubBus(partPin, to.start, to.width);
     }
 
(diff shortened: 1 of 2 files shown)


simulator/src/chip/chip.tsx
```diff
@@ -414,7 +414,7 @@ export class Chip {
     }
 
     // Wrap the chipPin in an OutBus when the part side is dimensioned
```
DavidSouther, line 416: Update the comment (the other one, the comment was right but the code was wrong!?)

DavidSouther (changes requested):

simulator/src/chip/chip.tsx
DavidSouther: The bodies are correct, but the conditions are backwards from the comment, which you've fixed. Ahh tricky

PR #232, opened 2024-02-19
Add edit mode to test panel

netalondon:


components/src/stores/chip.store.ts
```diff
@@ -334,13 +340,14 @@ export function makeChipStore(
...
         );
+        invalid = true;
         dispatch.current({
           action: "updateChip",
           payload: { invalid: true, error },
         });
         return;
       }
-      setStatus(`HDL code: No syntax errors`);
+      invalid ||= false;
```
DavidSouther, line 350: Does this do anything?

If invalid is false, it updates to false
If invalid is true, it does not update

components/src/stores/chip.store.ts
```diff
@@ -452,13 +459,15 @@ export function makeChipStore(
...
-        setStatus(`Failed to parse test`);
+        setStatus(`Failed to parse test ${Err(tst).message}`);
+        invalid = true;
         return false;
       }
 
       test = ChipTest.from(Ok(tst)).with(chip).reset();
       test.setFileSystem(fs);
       dispatch.current({ action: "updateTestStep" });
+      invalid ||= false;
```
DavidSouther, line 470: Noop

DavidSouther (approved):

DavidSouther (approved):

PR #236, opened 2024-03-05
Vm Emulator UI changes

netalondon:


simulator/src/cpu/memory.ts
```diff
@@ -106,7 +121,9 @@ export class Memory implements MemoryAdapter {
 
   loadBytes(bytes: number[], offset?: number): void {
     this.memory.set(new Int16Array(bytes), offset);
+    this.changed.fill(true, 0, bytes.length);
```
DavidSouther, line 124: Can the changed fills be adjacent? If so, I think they'd go better together. Regardless, a comment would be helpful. (It took me a few seconds to figure out what this was doing)

DavidSouther (changes requested):

DavidSouther (approved):

PR #239, opened 2024-03-12
Improve asm errors

netalondon:


components/src/chips/memory.tsx
```diff
@@ -197,8 +197,13 @@ export const Memory = ({
       : file.name.endsWith("asm")
       ? loadAsm
       : loadBlob;
-    const bytes = await loader(source);
-    memory.loadBytes(bytes);
+    try {
+      const bytes = await loader(source);
+      memory.loadBytes(bytes);
+    } catch (e) {
+      setStatus(`Error loading file: ${(e as Error).message}`);
```
DavidSouther, line 204: "file" is a bit overloaded, is there a more appropriate term we could use that's specific to this component? `Error loading memory` or `error loading memory from ${file.name}` or something?

DavidSouther (changes requested):

DavidSouther (approved):

PR #242, opened 2024-03-17
Hardware simulator and CPU emulator UI changes

netalondon:
A few more changes as requested by Shimon

### DIFF
--- a/components/package.json
+++ b/components/package.json
@@ -42,8 +42,8 @@
     "typescript": "^4.9.4"
   },
   "scripts": {
-    "copy": "rm -rf build/public && cp -r src/public/ build/public/",
-    "build": "tsc && npm run copy",
+    "build": "tsc",
+    "postbuild": "rm -rf build/public && cp -r src/public/ build/public/",
     "test": "react-scripts test"
   },
   "jest": {
--- a/components/package.json
+++ b/components/package.json
@@ -42,7 +42,8 @@
     "typescript": "^4.9.4"
   },
   "scripts": {
-    "build": "tsc",
+    "copy": "rm -rf build/public && cp -r src/public/ build/public/",
+    "build": "tsc && npm run copy",
     "test": "react-scripts test"
   },
   "jest": {
--- a/components/src/chips/alu.tsx
+++ b/components/src/chips/alu.tsx
@@ -19,8 +19,8 @@ export const ALUComponent = ({
 }) => (
   <div className="alu">
     <span>ALU</span>
-    <svg width="250" height="250" version="1.1" id="svg3">
-      <defs id="defs3">
+    <svg width="250" height="250" version="1.1">
+      <defs>
         <rect
           x="34.442518"
           y="54.335354"
@@ -28,7 +28,7 @@ export const ALUComponent = ({
           height="20.780869"
         />
       </defs>
-      <g id="g3">
+      <g>
         <polygon
           points="70,10 180,85 180,165 70,240 70,135 90,125 70,115"
           stroke="#000"
@@ -70,17 +70,17 @@ export const ALUComponent = ({
         >
           {COMMANDS_ALU.op[op] ?? "(??)"}
         </text>
-        <g id="g8">
+        <g>
           <path stroke="black" d="M 6,67.52217 H 68.675994" />
           <path stroke="black" d="M 68.479388,67.746136 60.290279,61.90711" />
           <path stroke="black" d="m 68.479388,67.40711 -8.189109,5.839026" />
         </g>
-        <g id="g13" transform="translate(0,115.5)">
+        <g transform="translate(0,115.5)">
           <path stroke="black" d="M 6,67.52217 H 68.675994" />
           <path d="M 68.479388,67.746136 60.290279,61.90711" stroke="black" />
           <path stroke="black" d="m 68.479388,67.40711 -8.189109,5.839026" />
         </g>
-        <g id="g16" transform="translate(176,57.5)">
+        <g transform="translate(176,57.5)">
           <path stroke="black" d="M 6,67.52217 H 68.675994" />
           <path stroke="black" d="M 68.479388,67.746136 60.290279,61.90711" />
           <path stroke="black" d="m 68.479388,67.40711 -8.189109,5.839026" />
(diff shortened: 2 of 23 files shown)


components/src/chips/alu.tsx
```diff
@@ -17,74 +17,116 @@ export const ALUComponent = ({
   out: number;
   flag: keyof typeof Flags;
 }) => (
-  <div>
+  <div style={{ fontSize: 20 }}>
```
DavidSouther, line 20: Use appropriate semantic styling, do not hard code style attributes in JSX.

components/src/chips/alu.tsx
```diff
@@ -17,74 +17,116 @@ export const ALUComponent = ({
...
+            d="M 6,67.52217 H 68.675994"
+            id="path11"
+          />
+          <path
+            style={{ stroke: "#000000" }}
+            d="M 68.479388,67.746136 60.290279,61.90711"
+            id="path12"
+          />
+          <path
+            style={{ stroke: "#000000" }}
```
DavidSouther, line 108: Prefer CSS rather than inline styles for SVG.

components/src/chips/alu.tsx
```diff
@@ -17,74 +17,116 @@ export const ALUComponent = ({
...
+            d="M 68.479388,67.746136 60.290279,61.90711"
+            id="path7"
+          />
+          <path
+            style={{ stroke: "#000000" }}
+            d="m 68.479388,67.40711 -8.189109,5.839026"
+            id="path8"
+          />
+        </g>
+        <g id="g13" transform="translate(0,115.5)">
```
DavidSouther, line 96: Are the IDs throughout this SVG necessary, or artifacts from an SVG editor?

components/src/chips/memory.tsx
```diff
@@ -296,7 +300,10 @@ export const Memory = forwardRef(
     });
 
     return (
-      <article className={`panel memory ${className ?? name}`}>
+      <article
+        className={`panel memory ${className ?? name}`}
+        style={{ overflowY: "auto" }}
```
DavidSouther, line 305: No inline styles.

components/src/pinout.tsx
```diff
@@ -284,7 +284,12 @@ const Pin = ({
         <>
           <div style={{ width: "1em" }} />
```
DavidSouther, line 285: You do it once as a hack and then it shows up everywhere as bad practice :) (Sorry for leaving this in as a bad example in the first place.)

components/src/pinout.tsx
```diff
@@ -284,7 +284,12 @@ const Pin = ({
...
           <button
-            style={{ maxWidth: "3em", margin: 0 }}
+            style={{
```
DavidSouther, line 287: Semantic CSS for this, please.

simulator/src/chip/builtins/computer/computer.tsx
```diff
@@ -76,8 +76,8 @@ export class Keyboard extends Chip implements KeyboardAdapter {
 }
 
 export class Memory extends ClockedChip {
-  private ram = new RAM16K();
```
DavidSouther, line 79: Why are you removing the `private` protections here and below?

web/src/pages/chip.tsx
```diff
@@ -196,23 +196,34 @@ export const Chip = () => {
             <option
               key={chip}
               value={chip}
-              style={isBuiltinOnly(chip) ? { color: "rgb(170, 170, 170)" } : {}}
+              style={
+                isBuiltinOnly(state.controls.project, chip)
+                  ? { color: "rgb(170, 170, 170)" }
```
DavidSouther, line 201: This color looks familiar, but has been repeated a dozen times in several formats. Can we consolidate and normalize? Preferably with a CSS custom property

web/src/pages/chip.tsx
```diff
@@ -370,6 +381,24 @@ export const Chip = () => {
     <TestPanel
       runner={runner}
       disabled={state.sim.invalid}
+      showLoad={false}
+      prefix={
+        state.controls.tests.length > 1 && (
```
DavidSouther, line 386: Does TestPanel correctly handle `prefix=false`? I think this might be better as 

```
state.controls.tests.length > 1 ? <select ...> : <></>;
```

DavidSouther (changes requested):

components/src/chips/alu.tsx
DavidSouther: The <g> still have IDs

web/src/pages/chip.tsx
DavidSouther: b0b0b0 in pin (176, but close enough for greyscale that it's effectively the same color visually).

components/src/pinout.tsx
DavidSouther: Hard coding styles tightly binds display and semantics, and adds more points for the developer to consider when implementing, troubleshooting, and maintaining the look and feel of the application. While just moving it to a class feels like moving deck chairs, it at least consolidates where to look for style settings. (CSS certainly could use better facilities to abstract styles.)

This StackOverflow post has a variety of good conversation on the topic https://stackoverflow.com/questions/2612483/whats-so-bad-about-in-line-css

DavidSouther (changes requested):
Still waiting on Screen size decisions from the email thread - want to pull that into a separate PR?

components/package.json
```diff
@@ -42,7 +42,8 @@
     "typescript": "^4.9.4"
   },
   "scripts": {
-    "build": "tsc",
+    "copy": "rm -rf build/public && cp -r src/public/ build/public/",
```
DavidSouther, line 45: This could just be `postbuild` - https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts

components/package.json
```diff
@@ -57,4 +58,4 @@
   "dependencies": {
     "@vscode/webview-ui-toolkit": "^1.2.1"
   }
-}
+}
```
DavidSouther, line 61: (nit) missing end \n

DavidSouther (changes requested):

DavidSouther (approved):

PR #248, opened 2024-03-25
Fix asm parsing bugs

netalondon:
Before the change, the assembler would allow for 2 instructions to be on the same line,
and would accept spaces between instruction parts, i.e `@ 2`

### DIFF
--- a/simulator/src/languages/asm.test.ts
+++ b/simulator/src/languages/asm.test.ts
@@ -10,7 +10,7 @@ describe("asm language", () => {
   });
 
   it("parses an A instruction to a label", () => {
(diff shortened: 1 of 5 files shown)


DavidSouther:
Aww I added that as a feature because I liked being able to write @AInst @CInst as one thing. I emailed Shimon to see if he's OK with that change in the spec.

DavidSouther:
Ohhh yeah that's a problem! There must be _some_ whitespace, but I found it
nice to have them on the same line. (And there should be no white space
between the @ and the value)
-------------------------
David Souther
http://davidsouther.com
(406) 545-9223

DavidSouther (approved):

PR #251, opened 2024-03-26
Allow DM,ADM as assembly destinations

netalondon:
This is the intended form as described in the book,
but the desktop simulator has it the other way around,
so Shimon proposed we support both forms.

### DIFF
--- a/simulator/src/languages/grammars/asm.ohm.js
+++ b/simulator/src/languages/grammars/asm.ohm.js
@@ -9,6 +9,8 @@ ASM <: Base {
   AInstruction = at (identifier | decNumber)
   CInstruction = assign? op jmp?
   
+  /* The book (figure 4.5) specifies DM and ADM as the correct forms for destination,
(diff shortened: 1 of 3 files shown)


simulator/src/languages/grammars/asm.ohm.js
```diff
@@ -10,10 +10,10 @@ ASM <: Base {
   CInstruction = assign? op jmp?
   
   assign = (
-      "AMD"
+      "AMD" | "ADM"
```
DavidSouther, line 13: Why not `MAD`? `DAM`? If we're doing down this road, shouldn't we allow all 6 permutations?

DavidSouther (changes requested):

simulator/src/languages/grammars/asm.ohm.js
DavidSouther: Good enough for me - could you add a comment to this effect in the .ohm file?

DavidSouther (approved):

PR #253, opened 2024-03-26
Load all files from local storage

netalondon:
UI changes:

1. Added a download button to the file picker.
Screenshot from 2024-03-31 15-18-34
For mobile:
Screenshot from 2024-03-31 15-19-02

2.Test panel now displays filename:
Screenshot from 2024-03-31 15-23-24

In the CPU Emulator, for cases in which there is more than one test file, we have a dropdown menu (like the hardware simulator):
Screenshot from 2024-03-31 15-23-13

3. VM Page: removed the "load file" button
vm


simulator/src/languages/asm.test.ts
```diff
@@ -345,22 +375,24 @@ describe("asm language", () => {
 
     // prettier-ignore
     expect(bin).toEqual([
-      0b0_000000000000000,    // @R0                     0x0000     0
```
DavidSouther, line 348: I chose this format for testing for a reason, because it clearly shows the structure of the Hack ASM. Could you keep this format for the test, and update the changed values?

web/src/pages/vm.tsx
```diff
@@ -127,43 +131,37 @@ const VM = () => {
...
+        // single file
         sources.push({
-          name: file.name.replace(".vm", ""),
-          content: await file.text(),
+          name: (path.split("/").pop() ?? path).replace(".vm", ""),
+          content: await fs.readFile(path),
         });
+      } else {
+        // folder
+        for (const file of await fs.scandir(path)) {
```
DavidSouther, line 148: I would collapse this a bit:

either
`for (const file of (await fs.scandir(path)).filter(f => f.isFile()) {`

or `if (file.isFile() && file.name.endsWith(".vm"))`

web/src/pages/vm.tsx
```diff
@@ -195,17 +193,17 @@ const VM = () => {
                 <Runbar
                   prefix={
                     <>
-                      <button
+                      {/* <button
```
DavidSouther, line 196: Delete it if we don't want it

web/src/pages/vm.tsx
```diff
@@ -219,7 +217,7 @@ const VM = () => {
                 />
               )}
             </div>
-            <input
+            {/* <input
```
DavidSouther, line 220: Delete

web/src/shell/file_select.tsx
```diff
@@ -77,20 +135,36 @@ export const FilePicker = () => {
...
+    downloadRef.current.download = chosen.split("/").pop() ?? chosen;
+    downloadRef.current.click();
+
+    URL.revokeObjectURL(url);
+  };
+
   return (
     <dialog open={filePicker.isOpen}>
-      <article style={{ width: "350px" }}>
+      <article style={{ width: "400px" }}>
```
DavidSouther, line 167: Move the constant to a css custom property so it has a name, and provide responsive breakpoint sizes

web/src/pages/asm.tsx
```diff
@@ -59,49 +61,19 @@ export const Asm = () => {
...
+    const path = await filePicker.select(".asm");
+    setStatus(LOADING);
+    requestAnimationFrame(async () => {
+      await actions.loadAsm(path);
+      setStatus("");
+    });
   };
 
   const { setStatus } = useContext(BaseContext);
-  const uploadFile = async (event: ChangeEvent<HTMLInputElement>) => {
```
DavidSouther, line 78: So we're dropping "Upload file from local computer"?

DavidSouther (changes requested):

web/src/shell/file_select.tsx
DavidSouther: On the smallest breakpoint (<576px) I expect making this dialog full-size will look and feel better. So I think we need two breakpoints for this width, small screen (default) at 100%+margin and non-mobile (@media min-width 576px) width: 400px+no-margin

DavidSouther (approved):

PR #254, opened 2024-03-26
Rename project folders (remove 0)

netalondon:


DavidSouther (commented):
This means we won't be able to `sort` lists of files.

DavidSouther:
Let's see if Shimon would be willing to do it on that side - it always bothered me as a student that they didn't sort nicely!

PR #255, opened 2024-03-26
Start CPU ROM as empty

netalondon:
Since implementing Mult.asm is part of project 4 it seems illogical that it will be loaded by default.


DavidSouther (approved):
I don't like the zero state being "nothing", but I also agree that starting with the project 4 solution isn't great... Starting with empty is probably better.

PR #256, opened 2024-03-27
Add assembly download button

netalondon:


DavidSouther (approved):

DavidSouther:
In the future, can we start including screenshots for UI changes?

netalondon:
> In the future, can we start including screenshots for UI changes?

Yes, not a problem

PR #274, opened 2024-04-24
Catch some HDL semantics errors that were previously ignored

netalondon:


simulator/src/chip/chip.tsx
```diff
@@ -387,7 +430,13 @@ export class Chip {
     return this.pins.emplace(from, minWidth);
   }
 
-  private wireOutPin(part: Chip, to: PinSide, from: PinSide) {
+  private wireOutPin(
+    part: Chip,
+    to: PinSide,
+    from: PinSide
+  ): Result<void, WireError> {
+    if (part.outs.get(to.name)) {
```
DavidSouther, line 438: Missing body?

DavidSouther (changes requested):

DavidSouther (approved):

PR #277, opened 2024-04-29
Feat: pwa manifest

DavidSouther:
Update manifest.json and generally make the PWA work.
Add service workers to load all files ahead of time.

To test this:

1. Run `npm run -w web serve-pwa` (requires installing latest node-modules, and having python3 available on your system path)
2. Visit localhost:8000/web-ide using Chrome or Edge
3. Click the "Install App" button
4. Test the app. Especially open web inspector, set Network to Offline, and ensure the app still functions as expected.


DavidSouther:
1. Done
2. To work offline, we will need to host them as PDFs in the app itself, but then we can't update them as quickly. So let's add an after deployment task to copy the PDFs from Google to the web/public directory, and update the URLs in the serviceWorker.js and the heading.tsx files.
3. They'll be used in the manifest.json "Screenshots" section, for users to preview while installing.

netalondon (approved):

PR #294, opened 2024-05-08
Change VM emulator to load from local file system

netalondon:


simulator/src/test/vmtst.ts
```diff
@@ -122,29 +122,29 @@ export class VMTest extends Test<VMTestInstruction> {
...
-          const file = await this.fs.readFile(
-            `${this.dir ? `${this.dir}/` : ""}${stat.name}`
-          );
-          files.push({ name: stat.name.replace(".vm", ""), content: file });
-        }
-      }
-      this.loadAction(files);
-    }
-  }
+  // override async load(filename?: string) {
```
DavidSouther, line 125: Remove if unused, but I think it's necessary?

DavidSouther (commented):
So the only way to get VM files is to upload them via drag and drop?

netalondon:
> So the only way to get VM files is to upload them via drag and drop?

We still have a file picker, but for the local file system instead of local storage.
If you have any reservations we can further discuss this with Shimon.

DavidSouther:
I was asking as a user guide question, not pushing back. How does this change the feature set? You can now only load local files?

Does your ideas for implementation change at all if we do the FileSystemAccess route instead?

netalondon:
> How does this change the feature set? You can now only load local files?

Yes

> Does your ideas for implementation change at all if we do the FileSystemAccess route instead?

Checking this with Shimon

PR #296, opened 2024-05-09
LocalStorage + FileSystemAccess file systems

DavidSouther:
This introduces a "Chained" FileSystemAdapter that allows using a FileSystemAccessFileSystemAdapter and a LocalStorageAdapter together. For reads, it will attempt to read from the FileSystemAccess, and use LocalStorage if it isn't found. For writes, it writes to both. To get this, you need to affirmatively choose a folder, you do this via the "Use Local FileSystem" settings option. (This needs to be selected every page load at this point.)


DavidSouther:
image


image


image

image

AndyPro720:
As of now, there's no way to upload from local fs on the published IDE right?

DavidSouther:
> As of now, there's no way to upload from local fs on the published IDE right?

Correct, and this PR is our first attempt at resolving that.

AndyPro720:
On a side note, is there any way to scale up/zoom the java VM emulator screen like on the IDE for ease of dev?

DavidSouther:
Yes, the x0, x1, and x2 buttons on the top-right of the screen:

image

image

AndyPro720:
@DavidSouther I was asking for a similar ease of use function/workaround on the Java VM emulator.  It's real handy on the IDE but I'm currently dev on local fs.

DavidSouther:
Ah, sorry, I got your question backwards. No, at this time, that is a feature that we've been able to add to the web version, and I don't anticipate we will be doing any updates to the Java version.

netalondon (approved):

PR #311, opened 2024-05-20
Add file system support to Jack compiler

netalondon:


components/src/stores/base/fs.ts
```diff
@@ -26,7 +26,7 @@ export class FileSystemAccessFileSystemAdapter implements FileSystemAdapter {
   ): Promise<FileSystemDirectoryHandle> {
     let folder = this.baseDir;
     const parts = path.split(SEP).slice(1);
-    for (const part of parts) {
+    for (const part of parts.filter((part) => part.trim() != "")) {
```
DavidSouther, line 29: There shouldn't be any empty directory pieces. If there are, the better way to handle it is a call to `Path::normalize`, but we don't have an easy to access copy of that for FS.

components/src/stores/compiler.store.ts
```diff
@@ -22,19 +23,41 @@ export type CompilerStoreDispatch = Dispatch<{
   payload?: unknown;
 }>;
 
+function classTemplate(name: string) {
+  return `class ${name} {
```
DavidSouther, line 27: I think this would look better as

    return `class ${name} {\n\n}\n`;

web/public/pico.min.css
DavidSouther, line 1: Revert this - there's a .prettierignore entry for it.

web/src/pages/compiler.tsx
```diff
@@ -220,3 +207,60 @@ export default Compiler;
...
+}) => {
+  const [value, setValue] = useState<string>();
+
+  return (
+    <dialog open={dialog.isOpen}>
+      <article>
+        <header>
+          <Trans>{title}</Trans>
+          <a
+            style={{ color: "rgba(0, 0, 0, 0)" }}
```
DavidSouther, line 232: Black transparent?

web/src/testing/index.tsx
```diff
@@ -8,7 +8,7 @@ import { i18n } from "@lingui/core";
 import { I18nProvider } from "@lingui/react";
 import { render, RenderOptions, RenderResult } from "@testing-library/react";
 import ue from "@testing-library/user-event";
-import { ReactElement } from "react";
+import { ReactElement, createContext } from "react";
```
DavidSouther, line 11: Remove

components/src/stores/compiler.store.ts
```diff
@@ -22,19 +23,41 @@ export type CompilerStoreDispatch = Dispatch<{
...
+
+    setFile(
+      state: CompilerPageState,
+      { name, content }: { name: string; content: string }
+    ) {
+      state.files[name] = content;
+      this.compile(state);
+    },
+
     setFiles(state: CompilerPageState, files: Record<string, string>) {
```
DavidSouther, line 54: Let's add a comment that `files` needs to be the full path, not the basename, of the file.

DavidSouther (changes requested):

components/src/stores/base/fs.ts
DavidSouther: In the mean time... move the `filter` to line 28 and we'll deal with it "later" :)

web/src/pages/compiler.tsx
netalondon: Oops, I guess that's what happens when you copy-paste. This was in settings as well.

Issue #327, opened 2024-05-27
[bug]: Order of ARegister and DRegister definitions matter in Project5/CPU


Apitheia:
### Program

Hardware Simulator

### Interface

Website (https://nand2tetris.github.io/web-ide)

### What happened?

Moving the DRegister to the bottom of the chip causes a timing issue for commands MD=D-1, AMD=D+A. The ALU outputs the wrong value until the following half clock step.

Moving the ARegister to the same location, but not the Mux16 and Or causes AMD=D+A error.

Moving the Or and Mux causes other errors in output that I haven't investigated.

```
CHIP CPU {
    
    IN  inM[16],         // M value input  (M = contents of RAM[A])
        instruction[16], // Instruction for execution
        reset;           // Signals whether to re-start the current
                         // program (reset==1) or continue executing
                         // the current program (reset==0).

    OUT outM[16],        // M value output
        writeM,          // Write to M? 
        addressM[15],    // Address in data memory (of M)
        pc[15];          // address of next instruction

    PARTS:
    
    Not(in=instruction[15], out=Ainstr);
    Or(a=instruction[15], b=false, out=Cinstr);
    Or(a=instruction[12], b=false, out=a);
    Or(a=instruction[11], b=false, out=c1);
    Or(a=instruction[10], b=false, out=c2);
    Or(a=instruction[9], b=false, out=c3);
    Or(a=instruction[8], b=false, out=c4);
    Or(a=instruction[7], b=false, out=c5);
    Or(a=instruction[6], b=false, out=c6);
    Or(a=instruction[5], b=false, out=d1);
    Or(a=instruction[4], b=false, out=d2);
    Or(a=instruction[3], b=false, out=d3);
    Or(a=instruction[2], b=false, out=j1);
    Or(a=instruction[1], b=false, out=j2);
    Or(a=instruction[0], b=false, out=j3);

    And(a=Cinstr, b=d1, out=writeA);
    And(a=Cinstr, b=d2, out=writeD);    
    And(a=Cinstr, b=d3, out=writeM);
    
    // move these
    Mux16(a=ALUout, b=instruction, sel=Ainstr, out=Ain);
    Or(a=writeA, b=Ainstr, out=loadA);
    ARegister(in=Ain, load=loadA, out=Aout, out[0..14]=addressM);
    DRegister(in=ALUout, load=writeD, out=Dout);
    //

    Mux16(a=Aout, b=inM, sel=a, out=y);
[...]

PR #328, opened 2024-05-28
Implement full topological sort

netalondon:
@DavidSouther turns out we did need a proper topological sort.
This is part of #327


DavidSouther (approved):

DavidSouther (changes requested):
(Clicked the wrong button)

simulator/src/chip/chip.test.ts
```diff
@@ -670,8 +671,43 @@ describe("Chip", () => {
...
+        {
+          from: { name: "a", start: 0, width: 16 },
+          to: { name: "out", start: 0, width: 16 },
+        },
+      ]);
+    }
+  }
+  const fooC = new FooC();
+  fooC.sortParts();
+  expect(fooC.parts).toEqual([fooC.register, fooC.inc16]);
```
DavidSouther, line 712: This currently passes on main without calling sortParts, so I don't think it shows the bug from #327

simulator/src/chip/builder.ts
```diff
@@ -220,6 +211,11 @@ class ChipBuilder {
     if (isErr(result)) {
       return result;
     }
+
+    this.chip.sortParts();
+    for (const part of this.chip.parts) {
+      part.subscribeToClock();
```
DavidSouther, line 217: This PR has gotten big enough it's trying to do two things - can we separate out the subscription updates from the sort updates?

simulator/src/chip/graph.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  sort(start: number) {
+    const visited = new Set<number>();
+    const visiting = new Set<number>();
+    const sorted: number[] = [];
+    const stack: NodeState[] = [{ node: start, isReturning: false }];
+
+    while (stack.length > 0) {
+      const { node, isReturning } = stack.pop()!;
```
DavidSouther, line 23: assertExists from `@davidsouther/jiffies` is a low overhead tool to not need the non-null assertion.

simulator/src/chip/graph.ts
```diff
@@ -0,0 +1,51 @@
...
+        sorted.push(node);
+      } else if (!visited.has(node)) {
+        // If we are visiting this node for the first time
+        if (visiting.has(node)) {
+          continue;
+        }
+        visiting.add(node);
+
+        // Re-push this node to handle it on return
+        stack.push({ node, isReturning: true });
```
DavidSouther, line 37: Oh very elegant way to handle this

DavidSouther (changes requested):

DavidSouther (commented):
Let me know when this is rebased and the minimal test is in place.

netalondon:
> Let me know when this is rebased and the minimal test is in place.

I updated the test and now I believe it fails on main.
Did I miss something?

simulator/src/chip/chip.test.ts
```diff
@@ -641,7 +641,10 @@ describe("Chip", () => {
     }
 
     const fooA = new FooA();
-    expect(fooA.parts).toEqual([fooA.notB, fooA.notA]);
```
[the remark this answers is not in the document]
DavidSouther: I updated the test in my branch to just look at the IDs, which is the important piece.

DavidSouther:
I opened #339 to keep the insertion sort, which I maintain is the neater and cleaner approach. But this PR was invaluable in tracking down a test case to make sure we got it right both here and for #327

Issue #330, opened 2024-05-31
Add "compiler" to the "Program" menu of this page. Also, change "Progarm" to "Tool"


Shimon-Schocken:
### Program

None

### Interface

Website (https://nand2tetris.github.io/web-ide)

### What feature are you proposing?

Feature description

### Do you want to try to add this feature?

- [ ] I want to try to add this feature!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #334, opened 2024-06-01
Temporarily hide "Upgrade fs" button in settings

netalondon:
Part of #330


web/src/shell/settings.tsx
```diff
@@ -153,7 +153,7 @@ export const Settings = () => {
                 <Trans>Files</Trans>
               </dt>
               <dd>
-                <button
+                {/* <button
```
DavidSouther, line 156: Prefer a flag and conditional rendering to commenting out.

DavidSouther (changes requested):

DavidSouther (approved):

PR #340, opened 2024-06-03
Capture CPUOutput in tock before updating CPUState from tock

### LINKED ISSUE #337: [bug]: Wrong value written to memory for MD=... or AMD=... commands
### Tool

CPU Emulator

### Interface

### Contact Details

_No response_

### What happened?

/web-ide/simulator/src/cpu/cpu.ts wrongly updates the D register during the tick. Therefore the following sequence will fail:
```
@0  
M=1 // init RAM[0]=1  
D=1  
MD=D+1 
```
D is 2 as expected, but RAM[0] will be 3 !

The CPU simulator acts as follows:
cpuTick(): D=D+1
cpuTock(): outM=[updatedD]+1=[originalD]+1+1 // new ALU calculation is triggered
Finally tick() writes the wrong outM to the memory

Any computation involving D with destinationD set, will result in the wrong value written to memory.
i.e. AMD=D+A, etc.

### Additional Comments

_No response_

### Code of Conduct

DavidSouther:
Closes #337

### DIFF
--- a/simulator/src/cpu/cpu.test.ts
+++ b/simulator/src/cpu/cpu.test.ts
@@ -131,7 +131,7 @@ describe("CPU", () => {
 
       const [output, outState] = cpu(input, inState);
 
-      expect(output).toEqual({ outM: 13, writeM: false, addressM: 0 });
+      expect(output).toEqual({ outM: 8, writeM: false, addressM: 0 });
       expect(outState).toEqual({
         A: 0,
         D: 8,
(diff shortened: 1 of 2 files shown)


netalondon:
Doesn't this conflict with CPU.tst which expects the current behavior?
Is the test wrong then?

axelkern:
If I understand the design of the CPU simulator correctly, the idea is that after the "tock" the CPU shows some intermediate state where no new instruction has been fetched, but the state of the CPU is otherwise updated and therefore the combinatorial outputs of the CPU show then a mix of old instruction with new register states. That is the reason of outM being wrong in this case.

This is somehow deviating from the hardware design of the Hack computer in my opinion, where the memory is implemented with a pure combinatorial Mux/Demux system without registering addresses. Consequently a new instruction will be immediately "fetched" when PC changes since it is just combinatorial. In a real world implementation, the best approach is probably to fetch the new instruction on the inverted clock signal, which is implicitly what the simulator does. In this case the outputs of the CPU would indeed show the above described intermediate state.

Proposal:
If this design should be kept as is, then the simplest way of implementing it from my point of view is to move the memory write into tick() into cpuTock() and do the write before A is updated. I think in this case the rest of the code can remain as it was and the tests should work.
```
// move this into cpuTock() before if (!bits.c) {...
if (writeM) {
      this.RAM.set(addressM, outM);
}

```

DavidSouther:
Closed in favor of #344

PR #355, opened 2024-06-10
Improve test panel diff view

netalondon:
New diff table tab, now separated from the output file tab:
Screenshot from 2024-06-10 10-06-41


DavidSouther:
This looks phenomenal

PR #389, opened 2024-06-26
Implement topological sort from Java verison

### LINKED ISSUE #360: [bug]: In Project 5: CPU, the order of the chips matters.
### Tool

Hardware Simulator

### Interface

### Contact Details

_No response_

### What happened?

If I put the registers above the ALU, I fail the test at time 19. If I put the ALU above the registers, I pass the test.

### Additional Comments

This one might actually be an issue with my understanding, not with the Web IDE. If this is desired behavior, and I don't understand how HDL works, then please accept my apologies.  

### Code of Conduct

netalondon:
Fixes #360

### DIFF
--- a/simulator/src/chip/builder.ts
+++ b/simulator/src/chip/builder.ts
@@ -213,10 +213,17 @@ class ChipBuilder {
     }
 
     // Reset clock order after wiring sub-pins
+    this.chip.clockedPins = new Set(
+      Array.from(this.chip.ins.entries())
+        .concat(Array.from(this.chip.outs.entries()))
+        .map((pin) => pin.name)
+        .filter((pin) => this.chip.isClockedPin(pin)),
+    );
     this.chip.sortParts();
     for (const part of this.chip.parts) {
       part.subscribeToClock();
     }
+    console.log(this.chip.parts.map((p) => p.name));
 
     return Ok(this.chip);
   }
--- a/simulator/src/chip/builder.ts
+++ b/simulator/src/chip/builder.ts
@@ -212,7 +212,6 @@ class ChipBuilder {
       return result;
     }
 
-    // Reset clock order after wiring sub-pins
     this.chip.clockedPins = new Set(
       Array.from(this.chip.ins.entries())
         .concat(Array.from(this.chip.outs.entries()))
@@ -220,10 +219,10 @@ class ChipBuilder {
         .filter((pin) => this.chip.isClockedPin(pin)),
     );
     this.chip.sortParts();
+    // Reset clock order after wiring sub-pins
     for (const part of this.chip.parts) {
       part.subscribeToClock();
     }
-    console.log(this.chip.parts.map((p) => p.name));
 
     return Ok(this.chip);
   }
--- a/simulator/src/chip/builder.ts
+++ b/simulator/src/chip/builder.ts
@@ -213,12 +213,13 @@ class ChipBuilder {
     }
 
     this.chip.clockedPins = new Set(
-      Array.from(this.chip.ins.entries())
-        .concat(Array.from(this.chip.outs.entries()))
+      [...this.chip.ins.entries(), ...this.chip.outs.entries()]
         .map((pin) => pin.name)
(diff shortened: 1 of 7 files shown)


simulator/src/chip/builder.ts
```diff
@@ -212,6 +212,13 @@ class ChipBuilder {
...
     }
 
+    this.chip.clockedPins = new Set(
+      Array.from(this.chip.ins.entries())
+        .concat(Array.from(this.chip.outs.entries()))
+        .map((pin) => pin.name)
+        .filter((pin) => this.chip.isClockedPin(pin)),
+    );
+    this.chip.sortParts();
     // Reset clock order after wiring sub-pins
```
DavidSouther, line 222: (nit) blank line between logical blocks

simulator/src/chip/builder.ts
```diff
@@ -212,6 +212,13 @@ class ChipBuilder {
       return result;
     }
 
+    this.chip.clockedPins = new Set(
+      Array.from(this.chip.ins.entries())
```
DavidSouther, line 216: This can be streamlined and (imo) reads better as:
```
[
  ...this.chips.ins.entries(),
  ...this.chips.outs.entries(),
]
.map((pin) => pin.name)
.filter((pin) => this.chip.isClockedPin(pin))
```

simulator/src/chip/builtins/computer/computer.tsx
```diff
@@ -324,6 +333,14 @@ export class Computer extends Chip {
       },
       { from: { name: "oldOutM", start: 0 }, to: { name: "out", start: 0 } },
     ]);
+
+    for (const pin of Array.from(this.ins.entries()).concat(
```
DavidSouther, line 337: For just one, not as bit a deal, but spread is slightly more idiomatic than Array.from. `[...this.ins.entries()]`

simulator/src/chip/chip.ts
```diff
@@ -419,11 +419,39 @@ export class Chip {
...
+        nodes.push(...(this.partToOuts.get(node) ?? []));
+      }
+    }
+
+    return false;
+  }
+
+  isClockedPin(pin: string) {
+    if (this.isInPin(pin)) {
+      return !Array.from(this.outs).some(([out, _]) =>
```
DavidSouther, line 442: Spread (and below)

simulator/src/chip/builder.ts
DavidSouther: (I know there's mixed usage, and I commented on it below. Behind the scenes, `Array.from` is looking for an object with a `.length` property that can be subscript-indexed, while spread uses the iterator protocol. Spread is more correct more often, so we should move to that as we're able.)

DavidSouther (changes requested):

simulator/src/chip/builder.ts
```diff
@@ -212,6 +212,14 @@ class ChipBuilder {
       return result;
     }
 
+    this.chip.clockedPins = new Set(
+      [...this.chip.ins.entries(), ...this.chip.outs.entries()]
+        .map((pin) => pin.name)
+        .filter((pin) => this.chip.isClockedPin(pin)),
```
DavidSouther, line 218: TIL this guidance from MDN:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map

> flatMap can be used as a way to add and remove items (modify the number of items) during a map.

```
this.chip.clockedPins = new Set(
  [...this.chip.ins.entries(), ...this.chip.outs.entries()]
  .flatMap(({name}) => this.chip.isClockedPin(name) ? [name] : [])
 );
 ```
 
But I'll be honest, it's not immediately obvious and it's not really saving that much in terms of work, so I'll leave it to you which you prefer aesthetically. I'm going to approve this PR & merge it, and if you like the flatMap, PRs are open ;)

DavidSouther (approved):

PR #390, opened 2024-06-26
Replace builtin CPU implementation

### LINKED ISSUE #376: [bug]: Chip part order still matters
### Tool

Hardware Simulator

### Interface

### Contact Details

_No response_

### What happened?

This implementation of the computer chip passes `ComputerAdd.tst`:
```
ROM32K(address=pc, out=instruction);
Memory(in=data, load=write, address=address, out=previousMout);
CPU(instruction=instruction, inM=previousMout, reset=reset, outM=data, writeM=write, addressM=address, pc=pc);
```

While this one fails:
```
ROM32K(address=pc, out=instruction);
CPU(instruction=instruction, inM=previousMout, reset=reset, outM=data, writeM=write, addressM=address, pc=pc);
Memory(in=data, load=write, address=address, out=previousMout);
```

(Only difference is switching the order of the last 2 lines)

### Additional Comments

_No response_

### Code of Conduct

netalondon:
Fixes #376.

I couldn't figure out what was wrong with the previous built-in CPU implementation, but at least this is a temporary fix.

### DIFF
--- a/components/src/pin_display.ts
+++ b/components/src/pin_display.ts
@@ -8,9 +8,8 @@ export class ChipDisplayInfo {
     if (BUILTIN_REGISTRY.has(chipName)) {
       const chip = assertExists(BUILTIN_REGISTRY.get(chipName)?.());
 
-      const pins = Array.from(chip.ins.entries()).concat(
-        Array.from(chip.outs.entries()),
-      );
+      const pins = [...chip.ins.entries(), ...chip.outs.entries()];
+
       for (const pin of pins) {
         this.signBehaviors.set(
           pin.name,
--- a/components/src/pin_display.ts
+++ b/components/src/pin_display.ts
@@ -1,25 +1,21 @@
-import { Ok, isOk } from "@davidsouther/jiffies/lib/esm/result.js";
-import {
-  REGISTRY as BUILTIN_REGISTRY,
-  getBuiltinChip,
-} from "@nand2tetris/simulator/chip/builtins/index.js";
+import { assertExists } from "@davidsouther/jiffies/lib/esm/assert.js";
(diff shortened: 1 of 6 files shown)


components/src/stores/chip.store.ts
```diff
@@ -597,15 +597,7 @@ export function makeChipStore(
       visualizationParameters: new Set(),
     };
 
-    const maybeChip = getBuiltinChip(controls.chipName);
```
netalondon, line 600: @DavidSouther any reason we default to the built-in chip here?
It gets overridden in `actions.initialize` anyway.

If not, could we remove/change the first line of the test at https://github.com/nand2tetris/web-ide/blob/main/components/src/stores/chip.store.test.ts#L103?

components/src/pin_display.ts
```diff
@@ -1,25 +1,21 @@
...
 
   public constructor(chipName: string, unsigned?: string[]) {
     if (BUILTIN_REGISTRY.has(chipName)) {
-      const chip = getBuiltinChip(chipName);
-      if (isOk(chip)) {
-        const pins = Array.from(Ok(chip).ins.entries()).concat(
-          Array.from(Ok(chip).outs.entries()),
+      const chip = assertExists(BUILTIN_REGISTRY.get(chipName)?.());
+
+      const pins = Array.from(chip.ins.entries()).concat(
```
DavidSouther, line 11: `[...chip.ins.entries(), ...chip.outs.entries]` (see discussion in the other PR)

components/src/stores/chip.store.ts
DavidSouther: Nope, just didn't see the holistic picture when I was working on it then. This is a good change.

DavidSouther (changes requested):

Issue #391, opened 2024-07-07
[bug]: Hardware simulator page becomes unresponsive


amcooper181:
### Tool

Hardware Simulator

### Interface

Website (https://nand2tetris.github.io/web-ide)

### Contact Details

amcooper181@gmail.com

### What happened?

Twice, the page has become unresponsive when working. The first time it occurred was when working on the Mux4Way16 in Project 1, and the second was when working on Bit in Project 3.

On both occasions, the website became slow to respond then stopped responding altogether. This occurred when adding code to the "Parts" section, not while running anything. Upon reloading, the page would not load and Google Chrome gave a "Page Unresponsive" warning (see image). 

<img width="1163" alt="Screenshot 2024-07-06 at 7 07 08 PM" src="https://github.com/nand2tetris/web-ide/assets/128009590/360c5ce8-83c3-40c2-91d1-43369c91dc23">

When using another browser, the issue was not present (but all progress was lost). Restarting the browser, as well as the computer, made no change. The only way to rectify the issue was to clear cookies for the IDE webpage, losing all progress. 

The issue occurred on:

-  Google Chrome version 126.0.6478.127
- Running on an Apple Macbook Pro with the M1 Pro processor running macOS Sonoma 14.0

### Do you want to try to fix this bug?

- [ ] I want to try to add this feature!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

DavidSouther:
From #392, "I tried clearing the chrome cache, and now the IDE runs perfectly fine on chrome"
From #391, "The only way to rectify the issue was to clear cookies"
From #357, "refreshing the page usually fixes the problem."

So there are differing reports of what will or will not fix the issue, it does seem to be generally related to storage access.

IvanKusturic:
<img width="1917" alt="Screenshot 2024-07-10 at 09 19 02" src="https://github.com/nand2tetris/web-ide/assets/85494415/97638453-5d09-4349-8b06-e5461be00ebc">

DavidSouther:
@IvanKusturic that looks unrelated to the page unresponsive issue, could you first reload the page and if the issue persists, open a new bug?

SeppahBaws:
Hi, I stumbled upon this myself as well, and I think it's because of an infinite loop that is triggered when you have a component have an output connect to its own input.

If I put this in the `Not` exercise in `Project1`, the page hangs.
```
CHIP Not {
  IN in;
  OUT out;
  PARTS:
[...]

PR #394, opened 2024-07-08
Fix bitmap editor url in service worker

netalondon:


netalondon:
@DavidSouther I think this could be the cause for #391 but I haven't been able to replicate it yet.
Should I close #391 in that case?

web/src/service-worker.ts
```diff
@@ -75,7 +75,7 @@ precacheAndRoute([
     url: "user_guide/compiler.pdf",
     revision: null,
   },
-  { url: "web-ide/bitmap_editor.html", revision: null },
```
DavidSouther, line 78: Shouldn't it be "/web-ide", like the static assets above?

DavidSouther (commented):
I'm not sure how this would resolve #391?

DavidSouther (approved):

PR #413, opened 2024-07-24
Add meta tags to disable page caching by browser

netalondon:
It seems like the IDE doesn't update even after deployment, and only updates when you clear the browser cache.
Hopefully this will solve the issue.
([Source](https://github.com/orgs/community/discussions/19713))

### DIFF
--- a/web/public/index.html
+++ b/web/public/index.html
@@ -5,6 +5,9 @@
     <title>NAND2Tetris</title>
     <meta name="description" content="NAND2Tetris Web IDE" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta http-equiv='cache-control' content='no-cache'> 
+    <meta http-equiv='expires' content='0'> 
+    <meta http-equiv='pragma' content='no-cache'>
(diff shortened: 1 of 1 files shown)


DavidSouther (approved):

web/public/index.html
```diff
@@ -5,6 +5,9 @@
     <title>NAND2Tetris</title>
     <meta name="description" content="NAND2Tetris Web IDE" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta http-equiv="cache-control" content="no-cache" />
+    <meta http-equiv="expires" content="0" />
+    <meta http-equiv="pragma" content="no-cache" />
```
DavidSouther, line 10: We don't need to pretend to handle http/1.0. The only equiv header this should need is `Cache-Control: no-cache` (note the "correct" (even if ignored by all user agents) capitalization).

DavidSouther (changes requested):

DavidSouther (approved):

PR #417, opened 2024-07-31
Align project file structure with software suite

netalondon:


simulator/src/chip/builtins/all.test.ts
```diff
@@ -25,7 +25,7 @@ describe("All Projects", () => {
     )("Builtin %s", async (chipName) => {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore
-      const ChipProject = ChipProjects[project]?.CHIPS[chipName];
+      const ChipProject = ChipProjects[project]?.CHIPS;
       let hdlFile: string = ChipProject?.[`${chipName}.hdl`];
```
DavidSouther, line 29: I think you can drop the ?. access as well?

DavidSouther (changes requested):

simulator/src/chip/builtins/all.test.ts
DavidSouther: Sorry, I meant below in the ChipProject?. parts

DavidSouther (approved):

DavidSouther (changes requested):
(wrong button)

DavidSouther (approved):

PR #419, opened 2024-07-31
Local FS UI 

### LINKED ISSUE #349: [bug]: Can not open a folder in Jack Compiler
### Tool

Jack Compiler

### Interface

### Contact Details

nickfegley@hotmail.com

### What happened?

I clicked the folder button in the Jack Compiler, expecting it to open the file dialog, but instead it did nothing. This is not an issue in the other views (CPU Emulator, Assembler, etc.) 

### Additional Comments

In the javascript console (Ctrl + Shift + i), the error is: `Uncaught (in promise) TypeError: window.showDirectoryPicker is not a function`

### Code of Conduct

### LINKED ISSUE #445: Jack compiler tooltip is not visible in firefox
### Tool

None

### Interface

### Contact Details

_No response_

### What happened?

Jack compiler tooltip is not visible in Firefox. For chrome it works. 
Firefox version 129.0.2
OS: Fedora 40

Is this expected?

### Additional Comments

_No response_

### Code of Conduct

netalondon:
Depends on #411 

Closes #349 
Closes #445


components/src/inline_edit.tsx
```diff
@@ -1,15 +1,14 @@
...
-import { Action } from "@nand2tetris/simulator/types.js";
 
 const Mode = { VIEW: 0, EDIT: 1 };
 
 export const InlineEdit = (props: {
   mode?: keyof typeof Mode;
   value: string;
   highlight: boolean;
-  onChange: Action<string>;
+  onChange: (value: string) => void;
```
DavidSouther, line 11: Reverted? (and a couple others)

components/src/stores/base.context.ts
```diff
@@ -19,84 +18,77 @@ import {
...
   removeLocalAdapterFromIndexedDB,
 } from "./base/indexDb.js";
-import { Action } from "@nand2tetris/simulator/types.js";
 
 export interface BaseContext {
   fs: FileSystem;
+  localFsRoot?: string;
   canUpgradeFs: boolean;
-  upgradeFs: (force?: boolean) => void;
+  upgradeFs: (force?: boolean) => Promise<void>;
```
DavidSouther, line 26: AsyncAction?

web/src/shell/file_select.tsx
```diff
@@ -28,10 +28,10 @@ export function useFilePicker() {
 
   const allowLocal = useRef(false);
 
-  const selected = useRef<(v: string | LocalFile) => void>();
+  const selected = useRef<(v: string | LocalFile[]) => void>();
```
DavidSouther, line 31: One string or an array of LocalFile? That feels a bit off, in that both will have a `.length` and be iterable, but iterating the string will get characters. Should this be `Array<string|LocalFile>`?

web/src/shell/file_select.tsx
```diff
@@ -125,31 +125,48 @@ function isFileValid(filename: string, validSuffixes: string[]) {
...
+function sortFiles(files: Stats[]) {
+  return files.sort((a, b) => {
+    const aIsNum = /^\d+$/.test(a.name);
+    const bIsNum = /^\d+$/.test(b.name);
+    if (aIsNum && !bIsNum) {
+      return -1;
+    } else if (!aIsNum && bIsNum) {
+      return 1;
+    } else if (aIsNum && bIsNum) {
+      return parseInt(a.name) - parseInt(b.name);
```
DavidSouther, line 137: Specify the radix when using parseInt (`04_foo` parses `04` as octal. I think actually this won't ever be "wrong" for _this_ comparison case, but I don't want to put the thought into proving that, so sticking with radix=10 will prevent surprises.

web/src/shell/file_select.tsx
```diff
@@ -230,38 +258,40 @@ export const FilePicker = () => {
...
-                  cd(file.name);
+          <div className="flex wrap files-container">
+            {fs.cwd() !== "/" && (
+              <FileEntry
+                stats={{
+                  isDirectory() {
+                    return true;
+                  },
+                  isFile() {
+                    return false;
```
DavidSouther, line 269: IMHO add this to the `files` array when you make it, rather than in the template.

web/src/shell/file_select.tsx
```diff
@@ -230,38 +258,40 @@ export const FilePicker = () => {
...
+                  name: "..",
+                }}
+                onDoubleClick={() => cd("..")}
+              />
+            )}
+            {files.map((file) => (
+              <FileEntry
+                key={file.name}
+                stats={file}
+                highlighted={file.name === chosen.split("/").pop()}
```
DavidSouther, line 280: `useMemo` to pre-compute `chosen.split().pop()` instead of in every loop.

web/src/shell/file_select.tsx
```diff
@@ -277,16 +307,18 @@ export const FilePicker = () => {
...
-            onClick={downloadFolder}
-            data-tooltip="Download all files in this folder into a zip"
-            disabled={chosen == "" || chosen.includes(".")}
-          >
-            Download
-          </button>
+          {!localFsRoot && (
+            <button
+              onClick={downloadFolder}
+              data-tooltip="Download all files in this folder into a zip"
```
DavidSouther, line 316: `t` macro for future translation.

web/src/shell/test_panel.tsx
DavidSouther, line 1: We can never edit tests through the web ide? What about writing tests for new chips? Should we tell users to do that outside the web ide?

web/src/shell/file_select.tsx
```diff
@@ -125,31 +125,48 @@ function isFileValid(filename: string, validSuffixes: string[]) {
     .reduce((p1, p2) => p1 || p2, false);
 }
 
+function sortFiles(files: Stats[]) {
+  return files.sort((a, b) => {
+    const aIsNum = /^\d+$/.test(a.name);
```
DavidSouther, line 130: Using parseInt allows non-digits after the initial numbers, so this doesn't need the `$` anchor, and then will "sort" `04_foo`  and `05_bar`.

web/src/pages/vm.tsx
```diff
@@ -129,35 +133,66 @@ const VM = () => {
...
+    let tstPath = "";
+    if (path.includes(".")) {
+      tstPath = path.replace(".vm", "VME.tst");
+    } else {
+      const name = (await fs.scandir(path)).find(
+        (entry) => entry.isFile() && entry.name.endsWith("VME.tst"),
+      )?.name;
+      tstPath = `${path}/${name}`;
+    }
+    console.log(tstPath);
```
DavidSouther, line 191: (nit) remove debugging logs

web/src/shell/test_panel.tsx
netalondon: With the compare-to change, users can no longer edit the compare files because they're loaded in runtime. 

It made sense to me then that we shouldn't allow editing the test scripts as well because you'd generally want to do both together, but I'm not sure...

What do you think?

components/src/stores/base.context.ts
netalondon: This will not allow calling `upgradeFs` with 0 agruments

web/src/shell/file_select.tsx
netalondon: As I see it you have 4 options (this is in case of no upgraded FS):
1. You pick a file from the browser storage - returns a path (string).
2. You pick a folder from the browser storage - returns a path (string) and we can later use `fs.scandir` on it.
3. You pick a file from the disk - returns a `LocalFile` object (can't just return the path because we'll have no way of accessing the file without the local file system API).
4. You pick a folder from the disk - returns a `LocalFile[]` for all of the files

So I don't think there's ever a reason to return a string array, but let me know if I'm missing something.

netalondon: I should add the we can differentiate between cases 1 and 2 by checking if the path includes a `.`

DavidSouther: > I should add the we can differentiate between cases 1 and 2 by checking if the path includes a .

Not entirely true, it is always possible to have a file without an extension.

Could you include all these details in a comment on the ref? (Or, create a custom type, comment it there, and use that here, like `type FileSelectionRef = string|LocalFile|LocalFile[]`

web/src/shell/test_panel.tsx
DavidSouther: This is on the "are we building an IDE or are we building a teaching tool" - we're on the teaching tool swing, so I think this is the right call for now. We can always change it later.

web/src/shell/file_select.tsx
netalondon: > Not entirely true, it is always possible to have a file without an extension.

Right, I'll add a field for that.

PR #423, opened 2024-08-12
Report error when chip parts have a loop

netalondon:
#401 didn't cover the case where the loop spanned over multiple parts.

For example:
```
CHIP Not {
	IN in;
	OUT out;

	PARTS:
	Nand(a=in, b=b, out=c);
	Nand(a=in, b=c, out=b);
}
```
Which as it stands still causes a crash.

### DIFF
--- a/simulator/src/chip/builder.test.ts
+++ b/simulator/src/chip/builder.test.ts
@@ -191,6 +191,26 @@ describe("Chip Builder", () => {
       throw new Error(display(e.message ?? e.shortMessage ?? e));
     }
   });
+
+  it("returns error for part loop", async () => {
+    try {
+      const chip = unwrap(
+        HDL.parse(`CHIP Not {
+        IN in;
+        OUT out;
+        PARTS:
+        Nand(a=in, b=b, out=c);
+        Nand(a=in, b=c, out=b);
+      }`),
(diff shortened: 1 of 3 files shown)


simulator/src/chip/builder.test.ts
```diff
@@ -191,6 +191,26 @@ describe("Chip Builder", () => {
...
+        PARTS:
+        Nand(a=in, b=b, out=c);
+        Nand(a=in, b=c, out=b);
+      }`),
+      );
+      const foo = await build(chip);
+      expect(foo).toBeErr();
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (e: any) {
```
DavidSouther, line 210: I thought we usually used `unknown` for this? Regardless, I would suggest using `toThrow` for this test. https://jestjs.io/docs/expect#tothrowerror

DavidSouther (approved):

simulator/src/chip/builder.test.ts
netalondon: But we're not expecting the `build` to throw anything, it returns a `Result` no?

It does look like the try-catch here is unnecessary, is that correct?

PR #430, opened 2024-08-16
Improve test diff colors for dark theme

happytomatoe:
This PR changes 
image
to
image

PR #432, opened 2024-08-17
Fix diff table colors in dark mode

lalit2005:
The red and green colors in the diff table lack contrast in dark mode. 
Made the text black so that it's visible in both - light and dark mode

Before:
image

After:
image


lalit2005:
similar to pr https://github.com/nand2tetris/web-ide/pull/430

PR #439, opened 2024-09-01
Allow colon in ASM symbols

### LINKED ISSUE #434: [bug]: Symbol error
### Tool

None

### Interface

### Contact Details

_No response_

### What happened?

When I try to use assembler for line
`@:R2`
I get an error
`Line 1: expected ".", a digit, "$", "_", or a letter`
while old java software compiles this with not issues. Was there a change in specification? If it is so would be so kind could you forward to me a new specification?

### Additional Comments

_No response_

### Code of Conduct

netalondon:
Fixes #434

### DIFF
--- a/simulator/src/languages/grammars/asm.ohm
+++ b/simulator/src/languages/grammars/asm.ohm
@@ -5,7 +5,6 @@ ASM <: Base {
   instruction = label|aInstruction|cInstruction
   intermediateInstruction = instruction space+
 
-  colon = ":"
   identifier := (letter|underscore|dot|dollar|colon) (alnum|underscore|dot|dollar|colon)*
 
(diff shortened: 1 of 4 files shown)


simulator/src/languages/grammars/asm.ohm
```diff
@@ -5,6 +5,9 @@ ASM <: Base {
   instruction = label|aInstruction|cInstruction
   intermediateInstruction = instruction space+
 
+  colon = ":"
```
DavidSouther, line 8: Should this go in base.ohm with the other character names?

DavidSouther (changes requested):

DavidSouther (approved):

PR #449, opened 2024-09-10
Fix issues with overriding OS classes

### LINKED ISSUE #438: [bug]: Compiler relying on built-in OS
### Tool

Jack Compiler

### Interface

### Contact Details

_No response_

### What happened?

1. Compiler reports an error when trying to call the init() functions of OS classes.
Example: Sys.jack
<pre><code>class Sys {
  function void init() {
    do Memory.init();
    return;
  }
}</code></pre>
Error message: Class Memory doesn't contain a subroutine init

2. Compiler ignores API of own implementations of OS classes.
Example: Sys.jack
<pre><code>class Sys {
  function void init() {
    do Sys.test();
    return;
  }
  function void test() {
    return;
  }
}</code></pre>

Error: Class Sys doesn't contain a subroutine test

3. Consequently the same error message also occurs when trying to call Sys.test from another class.
4. The above tests are working when using non-OS class names

It seems to be, that OS classes get a special treatment relying on built-in structures. First of all, the internally defined API doesn't seem to be complete (i.e. init() missing). More importantly I would say this behavior is problematic and is taking away implementation flexibility from users. Ultimately it results in not compiling correct code.

I presume, that the idea was to ensure that OS classes are in line with the API specs, however the restrictions go far beyond this. It would be enough to prevent attempts to (re-)define reserved OS functions not in line with the API, i.e. trying to define a function Math.divide (int a, int b, int c);

### Additional Comments

_No response_

### Code of Conduct

netalondon:
Fixes #438

### DIFF
--- a/simulator/src/jack/compiler.ts
+++ b/simulator/src/jack/compiler.ts
@@ -291,11 +291,31 @@ export class Compiler {
     }
   }
 
-  compileSubroutineDec(subroutine: Subroutine) {
+  validateSubroutineDec(subroutine: Subroutine) {
     this.validateReturnType(
       subroutine.returnType.value,
       subroutine.returnType.span,
     );
+
+    if (isOsClass(this.className)) {
+      const builtin = VM_BUILTINS[`${this.className}.${subroutine.name.value}`];
+      const argTypes = subroutine.parameters.map((arg) => arg.type.value);
+
+      if (
+        builtin &&
+        (builtin.args.length != argTypes.length ||
+          builtin.args.some((arg, index) => arg != argTypes[index]) ||
+          builtin.returnType != subroutine.returnType.value)
+      ) {
+        throw createError(
+          `OS subroutine ${this.className}.${subroutine.name.value} must follow the interface ${builtin.returnType} ${subroutine.name.value}(${builtin.args.join(",")})`,
+        );
+      }
+    }
+  }
+
+  compileSubroutineDec(subroutine: Subroutine) {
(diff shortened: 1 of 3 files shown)


simulator/src/jack/compiler.ts
```diff
@@ -291,11 +291,31 @@ export class Compiler {
...
       subroutine.returnType.span,
     );
+
+    if (isOsClass(this.className)) {
+      const builtin = VM_BUILTINS[`${this.className}.${subroutine.name.value}`];
+      const argTypes = subroutine.parameters.map((arg) => arg.type.value);
+
+      if (
+        builtin &&
+        (builtin.args.length != argTypes.length ||
```
DavidSouther, line 306: This feels like maybe it should be a function in vm/builtins.ts?

```
if (!builtinSubtype(builtin, argTypes)) {
  throw createError(`OS subroutine... ${makeBuiltinInterface(builtin)}`);
}
```

(names of methods are up to you)

simulator/src/vm/builtins.ts
```diff
@@ -26,7 +27,8 @@ export const VM_BUILTINS: Record<string, VmBuiltin> = {
       const [a, b] = getArgs(memory, 2);
       return (a * b) & 0xffff;
     },
-    nArgs: 2,
+    args: ["int", "int"],
```
DavidSouther, line 30: Love it

DavidSouther (changes requested):

DavidSouther (approved):

PR #459, opened 2024-09-16
Add compile jack comand to cli

happytomatoe:
As I am not using web-ide jack editor because it's slow, I've decided to add cli command to invoke Jack compiler directly

Screenshot from 2024-09-16 17-03-14

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import path, { dirname, parse, resolve } from "path";
-import yargs, { boolean } from "yargs";
+import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
 import { main } from "./grading.js";
 import { testRunner } from "./testrunner.js";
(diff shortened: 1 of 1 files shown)


DavidSouther:
@happytomatoe thank you so much for this PR! Definitely something we've had as a wishlist item while our effort has been focused on the web IDE. For your purposes, you might also look into a "CLI runner", that provides a CLI or (lightweight) windowed environment for Screen + Keyboard. I'll have a review for you later today. It looks like you can start by running `npm run ci` locally to ensure your CI will pass, and `npm run format ; npm run fix` to apply automatic formatting and lint fixes.

cli/src/index.ts
```diff
@@ -74,6 +78,96 @@ yargs(hideBin(process.argv))
...
+          await fs.writeFile(outpath, compiled);
+        } else {
+          if (!error) {
+            console.error("Compilation failed\n");
+          }
+          console.error(
+            Colors.Red +
+              compiled.message.replace(
+                /Line\s([\d]+):/g,
+                name + ".jack:$1" + Colors.Reset,
```
happytomatoe, line 158: FYI - I am using vscode and if  the message in this format one can ctr+click to go that line

happytomatoe:
@DavidSouther Could you elaborate on CLI runner? Currently in` cli/src/index.ts `I see that run command only accepts` .tst `files

cli/src/index.ts
DavidSouther: Nice!

DavidSouther:
>  I see that run command only accepts`.tst`files

Yes; this would be an entirely new project. `nand2tetris run file.tst` would still run a .tst file as expected. This new mode might be `nand2tetris exec myprog.hack`, where `myprog.hack` is the final compiled .jack -> vm -> hack binary file. This mode would then have the screen and keyboard interaction. Maybe we should start a new thread to discuss that!

DavidSouther:
Thank  you @happytomatoe for your first PR!

PR #460, opened 2024-09-16
Fix compiler page offset when having too much tabs

### LINKED ISSUE #451: [bug]: Jack compiler - page slides after using a slider and clicking on a tab when there are a lot of files
### Tool

None

### Interface

None

### Contact Details

_No response_

### What happened?

I have many files. When I use horizontal slider and click on last tab page slides
Screenshot from 2024-09-11 09-41-27

### Additional Comments

_No response_

### Code of Conduct

netalondon:
Fixes #451 

the line ` position: absolute !important;` caused the issue.
`disply: none;` achieves the same effect.

### DIFF
--- a/web/src/shell/tab.scss
+++ b/web/src/shell/tab.scss
@@ -63,8 +63,8 @@ label[role="tab"],
 }
 
 [role="tab"] [type="radio"] {
-  /* Leave the radio button focusable, but not visible. Copied from .screen-reader-text */
-  display: none;
+  opacity: 0; // keep the radio button focusable, but not visible, to allow switching tabs with the arrow keys
+  margin-inline: -7px; // offset the gap created by the invisible radio button (7px value was adjusted experimentally)
 }
 
 [role="tab"][aria-selected="true"] {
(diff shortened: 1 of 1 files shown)


DavidSouther:
Unfortunately, this breaks keyboard interactions. The current form allows keyboard focus on the tabs to use the arrow keys to move tabs.

netalondon:
We can do `opacity: 0;` instead, but this leaves a small gap after the label.

before:
Screenshot from 2024-09-16 18-04-16

after:
Screenshot from 2024-09-16 18-04-22

Is this good enough?

DavidSouther:
How does it look by removing the `position: absolute` and setting `opacity: 0; margin-inline: -1px;`?

netalondon:
> How does it look by removing the `position: absolute` and setting `opacity: 0; margin-inline: -1px;`?

There's still a gap but smaller:
Screenshot from 2024-09-16 19-01-19


`margin-inline: -7px;` looks most similar to the initial version.
Should I just do that?

DavidSouther:
It makes me a tiny bit uncomfortable, but just add a comment explaining why it's that way.

DavidSouther (approved):

PR #465, opened 2024-09-16
Add class name as prefix to VM labels in the compiler

### LINKED ISSUE #463: [bug]: VMEmulator is running with duplicated labels
### Tool

None

### Interface

None

### Contact Details

_No response_

### What happened?

If we are compiling 2+ jack files with at least 1 label in each, in VM code we will have L0 label duplicated. Is this expected? Seem like it shouldn't run at all. 
Why am I asking? I am trying to use vm translator for compiled files and I get duplicated label exception. Probably I will need to provide a hotfix for this one. Probably prepending file name to a label, in some format, will suffice. 

### Additional Comments

_No response_

### Code of Conduct

netalondon:
Fixes #463

### DIFF
--- a/simulator/src/jack/compiler.ts
+++ b/simulator/src/jack/compiler.ts
@@ -207,7 +207,7 @@ export class Compiler {
   }
 
   getLabel() {
-    const label = `${this.className}_L${this.labelNum}`;
+    const label = `${this.className}_${this.labelNum}`;
     this.labelNum += 1;
     return label;
   }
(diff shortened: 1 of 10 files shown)


simulator/src/jack/compiler.ts
```diff
@@ -207,7 +207,7 @@ export class Compiler {
   }
 
   getLabel() {
-    const label = `L${this.labelNum}`;
+    const label = `${this.className}_L${this.labelNum}`;
```
happytomatoe, line 210:
```suggestion
    const label = `${this.className}_${this.labelNum}`;
```

PR #467, opened 2024-09-17
Fix symbol table text color for dark theme

netalondon:
before:
Screenshot from 2024-09-17 10-30-32
after:
Screenshot from 2024-09-17 10-30-08


DavidSouther:
It's still a bit dark... could you contract following the steps in with https://developers.google.com/codelabs/devtools-cvd#3 and increase the brightness on the text until it's better visible?

DavidSouther (approved):

PR #473, opened 2024-09-17
Feat: breakpoints

happytomatoe:
TODO:
- [x] Fix bug when adding breakpoint causes editor to scroll up
- [ ] Add breakpoints to CPU Emulator
- [ ] Add breakpoints to test runner 
- [ ] Implement change when breakpoints should be cleared when a user uploads a file
- [ ] Add empty breakpoint - When a user presses on a function declaration simulation won't stop there - Won't do
- [ ] Handle a case when a user changes the code. In some cases simulator won't stop on a breakpoint - Probably won't do also
- [ ] Tests?

### DIFF
--- a/components/src/runbar.tsx
+++ b/components/src/runbar.tsx
@@ -1,5 +1,5 @@
 import { Timer } from "@nand2tetris/simulator/timer.js";
-import { ChangeEvent, ReactNode, useEffect, useRef } from "react";
+import { ChangeEvent, MutableRefObject, ReactNode, useEffect, useRef } from "react";
 import { useStateInitializer } from "./react.js";
 import { useTimer } from "./timer.js";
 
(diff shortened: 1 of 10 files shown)


happytomatoe:
Looks like speed is changed after adding a breakpoint in runbar

happytomatoe:
As it turned out that - as timer gets recreated, runbar also gets recreated. And after runbar creating it doesn't set the speed. So basically when I set a breakpoint timer uses default speed(1000). I've added a hotfix so that runbar sets the speed after being created

happytomatoe:
Would be great if someone could go over components as I see that they are recreated 3-5 times

happytomatoe:
I don't have time for this now

happytomatoe:
I am building the wrong feature. Old java CPU emulator has what is called watchpoint in eclipse. [VSCode doesn' have this feature](https://github.com/microsoft/vscode-cpptools/issues/294). I didn't even know that this exists. What a smart person built this feature! I won't continues building breakpoints feature. I don't think that people need it.
Currently I am trying to solve a specific issue - when I try to run a game, written in Jack, inside CPU emulator stack pointer goes into negative values. Watchpoint is an amazing tool to troubleshoot this.

PR #478, opened 2024-09-19
Add watch script

RomanLukash340:


build-watch.js
DavidSouther, line 1: This build-watch file seems like overkill - the set of projects is well known, and simply listing each one should be fine.

1. Add a `watch` target in each package (components, projects, simulator) that does `npm run build -- --watch`
2. Add a `watch` target in the root, which calls those directly using concurrently: `conc \"npm run -w components watch\" \"npm run -w projects watch\" \"npm run -w simulator watch\"`

DavidSouther (changes requested):

build-watch.js
happytomatoe: Changed

package.json
```diff
@@ -57,4 +56,4 @@
     "cross-env": "^7.0.3",
     "shx": "^0.3.4"
   }
-}
+}
```
DavidSouther, line 59: (nit) ending newline got lost

package.json
```diff
@@ -28,8 +28,7 @@
     "prebuild": "npm run check",
     "build": "cross-env CI=true npm run build -w projects && cross-env CI=true npm run build -w runner && cross-env CI=true npm run build -w simulator && cross-env CI=true npm run build -w cli && cross-env CI=true npm run build -w components && cross-env CI=true npm run build -w web && cross-env CI=true npm run build -w extension",
     "build:extension": "cross-env CI=true npm run build -w extension",
-    "watch": "node build-watch.js",
-    
+    "watch": "concurrently -k --restart-tries 3 \"npm run build -w components -- -w \" \"npm run build -w projects -- -w\" \"npm run build -w runner -- -w\" \"npm run build -w simulator -- -w\"   ",
```
DavidSouther, line 31: Could you clean up the whitespace in here a bit? Sometimes there's extra spaces before and after the \", sometimes not.

happytomatoe: if you mean at then end - done

package.json
happytomatoe: added newline

DavidSouther (approved):

Issue #447, opened 2024-09-08
[improvement]: Jack compiler is lagging with big files  


happytomatoe:
### Tool

Jack Compiler

### Interface

Website (https://nand2tetris.github.io/web-ide)

### What feature are you proposing?

Hi! 
I am writing a game in Jack and currently I have 7 files with 7238 lines. 
```
~nand2tetris/jack/dino/src$ wc -l *
   574 Cactus.jack
   142 Cloud.jack
   707 Dino.jack
    25 Main.jack
    10 MathCustom.jack
    14 ScreenCustom.jack
  5766 Trail.jack
  7238 total
```

After every key stroke I need to wait about a considerable amount of time before I can continue.
![Screenshot from 2024-09-08 09-29-34](https://github.com/user-attachments/assets/4002228d-ef6f-44f8-89bf-1fc72b954d3f)

Chrome shows this after I press run
![Screenshot from 2024-09-08 09-39-12](https://github.com/user-attachments/assets/2d44842a-907d-4267-8ff3-1d96d60decbd)

How to recreate this?
I guess if you will create a lot of Jack code that will do or you can use the code from my repo. 

Why my files are so big? 
I've created an utility that converts images into jack code and some of the sprites are quite long (like trail width>2000 pixels) and I am drawing each pixel individually. 

qq - Is this request has low priority? Additional question - do you happen to know if Java tool is still maintained?

### Do you want to try to add this feature?

- [ ] I want to try to add this feature!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

happytomatoe:
Please also note that VMEmulator from Java tool fails because program counter gets into negative values. I forgot that a program cannot have more than 2^15 lines. Probably even less for vm code as it will be translated into assembly where this program counter also exists.

happytomatoe:
I guess the error messages for the jack compiler in the old version are better. Please feel free to close this issue.

DavidSouther:
@happytomatoe thank you for your issue report. 

We will certainly leave it open to track possible performance improvements for large Jack projects, and thank you for the link to your repo so that we have some to test against! 

PS The java tool is generally unmaintained and released "as is" in its current state - while the code exists in several places online, bringing it to "modern" standards was considered to be less important than this web-based interface.

PR #483, opened 2024-10-03
Feat/compiler

happytomatoe:
Improved Jack compiler. Underneath the covers it uses ANTLR
List of improvements:
- Added more error messages  
- Closes  [Jack compiler is lagging with big file](https://github.com/nand2tetris/web-ide/issues/447). Ofc we can speed it up by calling compiler only once. As you can see that 1 key stroke calls compiler 2 times on screenshot.
Screenshot from 2024-10-03 17-39-47

Cons:
- I've removed the interfaces for built in classes. Though we don't have a validation rule to check if you supplying the right arguments(type) when calling OS classes
- If we want to use advanced ANTLR feature better to switch to JS target 

Don't get scared because it says 13k LOC. Most of it are tests and test files. The src changes is around 1k LOC. This doesn't include the generated files.

You can check it out on https://happytomatoe.github.io/web-ide/compiler

TODO:
- [X] Test lexer errors
- [X] Test parser
- [x] Test validations
- [x] Fix error message if class name doesn't match file name 
- [x] Validation doesn't work for empty function without return 
- [x] Fix bug connected to an error when playing Pong


simulator/package.json
```diff
@@ -22,6 +22,8 @@
     "@nand2tetris/projects": "^1.0.0",
     "@nand2tetris/runner": "^1.0.0",
     "@types/node": "^20.14.2",
+    "antlr4": "^4.13.2",
+    "assert": "^2.1.0",
```
happytomatoe, line 26: this is needed by ANTLR. I don't know why. They have support for many languages. Typescript is not their best one.

simulator/src/jack/builtins.ts
```diff
@@ -0,0 +1,93 @@
...
+export const builtInTypes = ["int", "boolean", "char"];
+interface Range {
+  min: number;
+  max: number;
+}
+export const intRange = { min: -32768, max: 32767 } as Range;
+//TODO: should we convert this to symbols?
+const builtInFunctionsToArgCount: Record<string, number> = {
+  "Array.dispose": 0,
+  //TODO: what is this?
```
happytomatoe, line 12: Would be great if someone could answer this

simulator/src/jack/listener/binder.listener.ts
```diff
@@ -0,0 +1,99 @@
...
+      );
+      return;
+    }
+    this.globalSymbolTable[className] = {} as GenericSymbol;
+    this.className = className;
+  };
+
+  override enterSubroutineDeclaration = (ctx: SubroutineDeclarationContext) => {
+    let subroutineType: SubroutineType;
+    if (ctx.subroutineType().CONSTRUCTOR() != null) {
```
happytomatoe, line 52: funny fact - Parser returns null if rule wasn't used.

simulator/src/jack/listener/validator.listener.ts
```diff
@@ -0,0 +1,583 @@
...
+      });
+  };
+  override enterSubroutineDeclaration = (ctx: SubroutineDeclarationContext) => {
+    if (ctx.subroutineType().CONSTRUCTOR() != null) {
+      this.subroutineType = SubroutineType.Constructor;
+      if (
+        ctx.subroutineDecWithoutType().subroutineReturnType().getText() !==
+        this.className
+      ) {
+        this.#addError(
```
happytomatoe, line 103: what convention do we use for private methods? Using ts private modifier doesn't make sense as we are using JS artifacts

simulator/src/jack/builtins.ts
DavidSouther: These are described in the second paragraph of section 12.3 of The Elements of Computing Systems.

> Init functions: Some OS classes use data structures that support the implementation of some of their subroutines. For each such OSClass, these data structures can be declared statically at the class level and initialized by a fucntion which, by convention, we call OSClass.init. The init functions are for internal purposes and are not documented in the OS API.

While they are not documented in the book, they are necessary to be defined for Jack programs that explicitly call them.

RomanLukash340:
Upd: I've added validate method to the compiler. Now after each keystroke time of execution is under 100ms for [many big files](https://github.com/nand2tetris/web-ide/issues/447).

simulator/src/jack/builtins.ts
RomanLukash340: Changed

.vscode/launch.json
```diff
@@ -34,6 +34,13 @@
       ],
       "outFiles": ["${workspaceFolder}/extension/build/test/**/*.js"],
       "preLaunchTask": "${defaultBuildTask}"
+    },
```
RomanLukash340, line 37: run web ide doesn't work for me(vscode doesn't stop at the breakpoint place ). That's why I've added additional

simulator/src/jack/resources/test/Average/Main.jack
```diff
@@ -0,0 +1,27 @@
+// This file is part of www.nand2tetris.org
```
RomanLukash340, line 1: Would gladly accept suggestions if this the wrong folder for these files

RomanLukash340:
qq - why do we write VM files to fs?

happytomatoe:
@DavidSouther @netalondon just FYI - I've created new repo for [jack language server protocol](https://github.com/happytomatoe/nand2tetris-jack-lsp), I would gladly contribute this to nand2tetris repo.
About this PR - I would rather have extract compiler into its' own package because of the new repo

DavidSouther:
Great work in here, I've been following your WIP as it goes.

We're happy to have the LSP in this repo, we just need to spend a bit thinking through where it will go, how to extend it to the other languages, and how to integrate it with the web, cli, and ide extensions.

Again, your use case as a "Jack developer" is relatively niche in the NAND2Tetris community. The vast bulk of work happens at the Hack level, and a little bit at the VM and assembly levels. Improvements to those will have the most impact for NAND2Tetris students.

So finding a migration path from the current state to your ANTLR based approach is a good project for quarter, if someone wanted to pick that up.

Are you ready for me to review this PR?

happytomatoe:
@DavidSouther  Please go ahead. There is a possibility that I can find some small bugs. I can create another PR, if you would like, to fix them.

DavidSouther:
>  As you can see that 1 key stroke calls compiler 2 times on screenshot.

In dev, yes; in production, it should still only be called once. That said, we can probably make better use of effects to ensure it happens only once either way.

simulator/src/jack/symbol.ts
```diff
@@ -0,0 +1,131 @@
...
+    [ScopeType.Argument]: [],
+    [ScopeType.Local]: [],
+  };
+  private static scopesLookupOrder = [
+    ScopeType.Local,
+    ScopeType.Argument,
+    ScopeType.This,
+    ScopeType.Static,
+  ];
+  lookup(name: string): VariableSymbol | undefined {
```
DavidSouther, line 80: (nit) here and elsewhere - one blank line between declarations

simulator/src/jack/listener/binder.listener.test.ts
```diff
@@ -0,0 +1,116 @@
+import fs from "fs";
```
DavidSouther, line 1: I know it's a bit more work, but please use the `fs` abstraction from Jiffies that's used elsewhere in the project. It allow much easier mocking and replacement later on.

simulator/src/jack/listener/binder.listener.ts
```diff
@@ -0,0 +1,99 @@
...
+  private subroutineVarsCount = 0;
+  private stopProcessingSubroutines = false;
+  private subroutineId = "";
+
+  override enterClassDeclaration = (ctx: ClassDeclarationContext) => {
+    const id = ctx.className()!.IDENTIFIER();
+    const className = id.getText();
+    if (this.globalSymbolTable[className] != undefined) {
+      const e = new DuplicatedClassError(
+        ctx.className()!.start.line,
```
DavidSouther, line 38: Maybe extract ctx.className one time?

simulator/src/jack/listener/binder.listener.ts
```diff
@@ -0,0 +1,99 @@
...
+  public globalSymbolTable: GlobalSymbolTable = structuredClone(builtInSymbols);
+  public className = "";
+  public errors: DuplicatedSubroutineError[] = [];
+  private subRoutineInfo: SubroutineInfo = {} as SubroutineInfo;
+  private subroutineVarsCount = 0;
+  private stopProcessingSubroutines = false;
+  private subroutineId = "";
+
+  override enterClassDeclaration = (ctx: ClassDeclarationContext) => {
+    const id = ctx.className()!.IDENTIFIER();
```
DavidSouther, line 34: Prefer the `assertExists` utility to the `!` type override operator.

simulator/src/jack/listener/error.listener.ts
```diff
@@ -0,0 +1,47 @@
...
+} from "antlr4";
+import { JackCompilerError, LexerOrParserError } from "../error.js";
+interface LexerNoViableAltException {
+  startIndex: number;
+}
+export class CustomErrorListener extends ErrorListener<any> {
+  public errors: JackCompilerError[] = [];
+
+  override syntaxError = (
+    recognizer: Recognizer<any>,
```
DavidSouther, line 16: No any, anywhere. `unknown` is acceptable.

simulator/src/jack/listener/validator.listener.test.ts
```diff
@@ -0,0 +1,840 @@
...
+            function void a(){
+                let b=1;
+                return;
+            }
+            }`,
+      UndeclaredVariableError,
+    );
+  });
+
+  test("call function - undeclared variable ", () => {
```
DavidSouther, line 110: It looks like most of these are just doing the same `testValidator` call, so you could probably use `test.each` for a parameterized or data-driven test. https://vitest.dev/api/#test-each

simulator/src/jack/listener/validator.listener.ts
```diff
@@ -0,0 +1,604 @@
+import { ParserRuleContext } from "antlr4";
+import { builtInTypes, intRange } from "../builtins.js";
+import {
```
DavidSouther, line 3: I would actually do a * import for these ones, and do `errors.FieldCantBeReferencedInFunction` or similar throughout this file.

simulator/src/jack/listener/validator.listener.ts
DavidSouther: Why do you need to block anyone else from calling addError? I don't know how anyone would get a reference to an instance of the validator outside the tree, so it doesn't seem particularly worthwhile to have these as private.

simulator/src/jack/listener/validator.listener.ts
```diff
@@ -0,0 +1,604 @@
...
+
+  override enterVarDeclaration = (ctx: VarDeclarationContext) => {
+    const type = ctx.varType().getText();
+    ctx.varNameInDeclaration_list().forEach((name) => {
+      this.#localSymbolTableAdd(ctx, ScopeType.Local, name.getText(), type);
+    });
+  };
+
+  /**
+   * Var name when doing some actions - do statement, let ... We have a different rule for a var name that is used in declaration
```
DavidSouther, line 180: I can't quite follow this comment. Please use complete English sentences in docblock comments.

simulator/src/jack/listener/validator.listener.ts
```diff
@@ -0,0 +1,604 @@
...
+        ),
+      );
+      this.stopProcessingErrorsInThisScope = true;
+    }
+  };
+  override enterRBrace = (ctx: RBraceContext) => {
+    this.stopProcessingErrorsInThisScope = false;
+  };
+  /**
+   * Control flow
```
DavidSouther, line 238: This isn't a particularly helpful comment - it's meant as a section separator, but it's written as a docblock attached to enterWhileStatement. I would suggest simply removing it.

simulator/src/jack/listener/validator.listener.ts
```diff
@@ -0,0 +1,604 @@
...
+    const varName = ctx.varName();
+    const constCtx = ctx.expression().constant();
+    //corresponding literal type check
+    if (
+      varName != null &&
+      constCtx != null &&
+      this.localSymbolTable.lookup(ctx.varName()!.getText()) &&
+      ctx.expression().constant()!.NULL_LITERAL() == null
+    ) {
+      const type = this.localSymbolTable.lookup(ctx.varName()!.getText())!.type;
```
DavidSouther, line 278: As elsewhere, `assertExists` in lieu of `!` assertions.

simulator/src/jack/listener/validator.listener.ts
```diff
@@ -0,0 +1,604 @@
...
+    if (this._returns) {
+      return this._returns;
+    } else if (this.right == undefined && this.left == undefined) {
+      return false;
+    } else if (this.right != undefined && this.left != undefined) {
+      return this.left.returns && this.right.returns;
+    } else if (this.left != undefined) {
+      return false;
+    } else {
+      throw new Error("Something went wrong - CFG has only right  subtree");
```
DavidSouther, line 555: (nit) 2 spaces

simulator/src/jack/listener/validator.listener.ts
```diff
@@ -0,0 +1,604 @@
...
+  }
+
+  public set returns(_returns: boolean) {
+    this._returns = _returns;
+  }
+  print() {
+    console.log("Branch returns value");
+    console.log(".");
+    console.log(this.printBT());
+  }
```
DavidSouther, line 566: Don't mix output into data classes. This can simply return the string, and allow the caller to decide how to handle the IO.

simulator/src/jack/listener/vm.writer.listener.ts
```diff
@@ -0,0 +1,286 @@
...
+  "<": "lt",
+  ">": "gt",
+  "=": "eq",
+};
+
+const unaryOperationToVmCmd: Record<string, string> = {
+  "-": "neg",
+  "~": "not",
+};
+/* eslint-disable  @typescript-eslint/no-non-null-assertion */
```
DavidSouther, line 43: Don't disable eslint.

simulator/src/jack/listener/vm.writer.listener.ts
```diff
@@ -0,0 +1,286 @@
...
+/**
+ * Transforms parse tree into VM code
+ */
+export class VMWriter extends JackParserListener {
+  public result = "";
+  private className = "";
+  private currentLabelInd = 0;
+  private localSymbolTable: LocalSymbolTable | undefined;
+  private afterEquals = false;
```
DavidSouther, line 52: I would maybe change this name to `inAssignmentRightHandSide`? I think that's how it's being used? It only becomes true when reviewing the assignment operator to end of statement, and Jack doesn't allow multiple assignment.

simulator/src/jack/listener/vm.writer.listener.ts
```diff
@@ -0,0 +1,286 @@
...
+  "~": "not",
+};
+/* eslint-disable  @typescript-eslint/no-non-null-assertion */
+/**
+ * Transforms parse tree into VM code
+ */
+export class VMWriter extends JackParserListener {
+  public result = "";
+  private className = "";
+  private currentLabelInd = 0;
```
DavidSouther, line 50: Prefer `Index` to `Ind`

simulator/src/jack/listener/vm.writer.listener.ts
```diff
@@ -0,0 +1,286 @@
...
+const unaryOperationToVmCmd: Record<string, string> = {
+  "-": "neg",
+  "~": "not",
+};
+/* eslint-disable  @typescript-eslint/no-non-null-assertion */
+/**
+ * Transforms parse tree into VM code
+ */
+export class VMWriter extends JackParserListener {
+  public result = "";
```
DavidSouther, line 48: I would suggest keeping a bit of structure here, and storing this not as the program text of the VM code, but instead using the VM language primitives. We can then use that serialization (or, load that directly into the VM emulator)

simulator/src/jack/compiler.ts
DavidSouther, line 1: Rather than deleting this file, could you create a second `antlr_compiler.ts` class, and we'll delete this one later?

DavidSouther (changes requested):
Great work on getting this updated compiler working! Love it big picture. I have a variety of comments - some are small nitpicks on style things that Prettier doesn't capture. Some are suggestions for using utility libraries and classes that the project uses consistently.  I have a few suggestions for improving a couple pieces of the design implementation, happy to have a longer conversation on those.

components/src/stores/compiler.store.ts
```diff
@@ -48,11 +50,15 @@ export function makeCompilerStore(
 
     setFile(
       state: CompilerPageState,
-      { name, content }: { name: string; content: string },
+      { name, content }: FileEntry,
     ) {
       state.files[name] = content;
       state.isCompiled = false;
-      this.compile(state);
+    },
```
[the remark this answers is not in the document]
happytomatoe: What's the rule here? Can we add this to prettier?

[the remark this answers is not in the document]
happytomatoe: Please ignore the piece about prettier

simulator/README.md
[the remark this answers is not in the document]
happytomatoe: Could you check if it looks ok now?

.gitignore
```diff
@@ -24,3 +24,5 @@ out
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+**/.vscode/numbered-bookmars.json
+**/.antlr
```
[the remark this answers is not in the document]
happytomatoe: Done

simulator/src/jack/listener/vm.writer.listener.ts
happytomatoe: Can I ask why? Antlr returns nulls  if the derivation didn't happen in some case(let's say we have static instead of field keyword) . In this file I need to assure compiler that after I've checked that some property is not null using "!". 
Should I just remove this and leave the compiler warnings?

simulator/src/jack/listener/vm.writer.listener.ts
happytomatoe: What are the vm language primitives? Do you mean the instructions array(array of strings)?

simulator/src/jack/compiler.ts
happytomatoe: sure

simulator/src/jack/listener/vm.writer.listener.ts
happytomatoe: Done

simulator/src/jack/listener/vm.writer.listener.ts
happytomatoe: Done

simulator/package.json
```diff
@@ -35,6 +36,10 @@
   },
   "scripts": {
     "build": "tsc",
-    "test": "jest --verbose"
+    "test": "jest --verbose",
+    "test-w": "jest --watchAll --verbose",
+    "test-jack": "npm run test -- -t \"Jack\"",
```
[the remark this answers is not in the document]
happytomatoe: Removed

simulator/src/jack/builtins.ts
```diff
@@ -0,0 +1,90 @@
+import { GenericSymbol, SubroutineType } from "./symbol.js";
+
+export const builtInTypes = ["int", "boolean", "char"];
+interface Range {
+  min: number;
+  max: number;
+}
+export const intRange = { min: -32768, max: 32767 } as Range;
```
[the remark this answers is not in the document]
happytomatoe: Done

simulator/src/jack/builtins.ts
```diff
@@ -0,0 +1,90 @@
...
+  "String.length": 0,
+  "String.new": 1,
+  "String.newLine": 0,
+  "String.setCharAt": 1,
+  "String.setInt": 1,
+  "Sys.error": 1,
+  "Sys.halt": 0,
+  "Sys.wait": 1,
+};
+const builtInClasses = [
```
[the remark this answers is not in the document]
happytomatoe: Done

simulator/src/jack/error.ts
```diff
@@ -0,0 +1,287 @@
+export interface Span {
```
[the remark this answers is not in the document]
happytomatoe: Changed

simulator/src/jack/error.ts
```diff
@@ -0,0 +1,287 @@
...
+  start: number;
+  end: number;
+  line: number;
+}
+export class JackCompilerError {
+  public span: Span;
```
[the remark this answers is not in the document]
happytomatoe: Thank you! Changed

simulator/src/jack/error.ts
```diff
@@ -0,0 +1,287 @@
...
+  }
+  public toString = (): string => {
+    return (
+      this.constructor.name +
+      `(${this.span.line}:${this.span.start}:${this.span.end} ${this.msg})`
+    );
+  };
+}
+
+export class LexerOrParserError extends JackCompilerError {}
```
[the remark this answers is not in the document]
happytomatoe: Yes, please provide code suggestion. From my standpoint I use this because it's easier to check in the tests the error type rather than message

simulator/src/jack/error.ts
```diff
@@ -0,0 +1,287 @@
...
+    subroutineName: string,
+    className?: string,
+  ) {
+    super(
+      line,
+      startInd,
+      endIndex,
+      `Can't find subroutine '${subroutineName}'${className ? ` in ${className}` : ""}`,
+    );
+    Object.setPrototypeOf(this, UnknownSubroutineCallError.prototype);
```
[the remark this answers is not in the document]
happytomatoe: Deleted. Previously this  class extended error type.

simulator/src/jack/listener/binder.listener.test.ts
happytomatoe: Sorry, that doesn't make sense to me as fs abstraction doesn't have synchronous methods to read a file or a directory.  Is there any other approach rather than using callback function (I mean .then method)?

simulator/src/jack/listener/binder.listener.ts
```diff
@@ -0,0 +1,99 @@
...
+} from "../symbol.js";
+import JackParserListener from "../generated/JackParserListener.js";
+
+const primitives = new Set(builtInTypes);
+export type Primitive = typeof primitives extends Set<infer S> ? S : never;
+/* eslint-disable  @typescript-eslint/no-non-null-assertion */
+/**
+ * Creates global symbol table that contains built-in functions and found classes and subroutines
+ */
+export class BinderListener extends JackParserListener {
```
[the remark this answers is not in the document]
happytomatoe: Because this concept is used for typescript. Please suggest better name

happytomatoe:
> > So finding a migration path from the current state to your ANTLR based approach is a good project for quarter, if someone wanted to pick that up.
> 
> @DavidSouther Can you create a new repo? I would provide the scaffolding along with the jack lsp implementation

Nvm. I don't want to spend my time on that.

components/src/stores/compiler.store.ts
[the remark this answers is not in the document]
DavidSouther: Yeah, unfortunately I don't think prettier supports it at this time. I believe Rome might? But I haven't looked into it closely.

Broadly, separate top-level and class-level function declarations by one blank line. Top level and class level properties and variables can be grouped without blank lines between them "so that it looks good" - I know that's not a great rule, but there's room for a bit of aesthetics. "Keep similar things together" is the principle there, but not super strict.

But for the rule - every function should have a blank line before it (or before its docblock, if it has a comment)

components/src/stores/compiler.store.ts
```diff
@@ -136,15 +152,29 @@ export function makeCompilerStore(
         await fs.writeFile(`${name}.jack`, content);
       }
     },
+    async writeNewFile(name: string, content?: string) {
```
DavidSouther, line 155: So, this one and `validate` should have a blank line before them, but `reset` and `compile` look good.

simulator/README.md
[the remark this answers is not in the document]
DavidSouther: The write up in [simulator/src/jack/README.md](https://github.com/nand2tetris/web-ide/pull/483/files/#diff-7212235467ce085d36f890e059c2b3082d4c1f581a429fcd9513888199c6408f) is great, thank you!

simulator/src/jack/error.ts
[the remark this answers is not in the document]
DavidSouther: What about this? https://github.com/happytomatoe/web-ide/pull/4

DavidSouther:
@happytomatoe I hope you're able to keep making progress on this, but I wanted to let you know that I'm unavailable for the coming week. I'll be back October 20th to review any changes.

components/src/stores/compiler.store.ts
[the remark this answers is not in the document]
happytomatoe: Is this what are we looking for?
https://stackoverflow.com/questions/52772097/how-to-add-new-line-after-and-before-method-declarations-in-classes-using-pretti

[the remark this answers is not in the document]
DavidSouther: Ah, it's a part of eslint, not prettier, that's probably why I couldn't find it.

simulator/src/jack/error.ts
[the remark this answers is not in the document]
DavidSouther: You're probably thinking of their `Diagnostic` interface, which does something similar albeit using Number for identifying the error code, rather than a string. Regardless of formatting and style, a class hierarchy is not the correct approach in TS. Happy to iterate the tagged union to something you find more readable, but it will not be class based.

simulator/src/jack/compiler.ts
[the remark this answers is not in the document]
DavidSouther: Looks good.

simulator/src/jack/listener/binder.listener.test.ts
happytomatoe: Thank you! Changed.

simulator/src/jack/listener/binder.listener.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      .map((file) => path.join(testFolder, file));
+    for (const filePath of files) {
+      const tree = parseJackFile(filePath);
+      globalSymbolsListener = listenToTheTree(tree, globalSymbolsListener);
+      // console.log("Symbols for " + path.basename(filePath) + ":", globalSymbolsListener.globalSymbolTable)
+    }
+    expect(globalSymbolsListener.globalSymbolTable).toEqual(expected);
+  });
+});
+function testBinder<T extends { name: string }>(
```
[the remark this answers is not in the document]
happytomatoe: Great! I didn't know that this could be done this way. Thank you!

simulator/src/jack/listener/binder.listener.ts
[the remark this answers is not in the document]
happytomatoe: Changed to JackGlobalSymbolTableListener

simulator/src/jack/listener/binder.listener.ts
happytomatoe: Done

simulator/src/jack/listener/binder.listener.ts
happytomatoe: Done

simulator/src/jack/listener/common.ts
```diff
@@ -0,0 +1,45 @@
+import { SubroutineIdContext } from "../generated/JackParser.js";
+import { LocalSymbolTable, VariableSymbol } from "../symbol.js";
+
+export interface CallTypeResult {
+  callType: CallType;
+  subroutineIdText: string;
+  symbol?: VariableSymbol;
+}
+
+export enum CallType {
```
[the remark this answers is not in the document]
happytomatoe: @DavidSouther Can I ask you why? I specially used int enum for this to be faster. I guess ANTLR uses the same strategy

simulator/src/jack/listener/binder.listener.ts
```diff
@@ -0,0 +1,99 @@
...
+  GenericSymbol,
+  GlobalSymbolTable,
+  SubroutineInfo,
+  SubroutineType,
+} from "../symbol.js";
+import JackParserListener from "../generated/JackParserListener.js";
+
+const primitives = new Set(builtInTypes);
+export type Primitive = typeof primitives extends Set<infer S> ? S : never;
+/* eslint-disable  @typescript-eslint/no-non-null-assertion */
```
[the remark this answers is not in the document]
happytomatoe: Done

simulator/src/jack/listener/error.listener.ts
happytomatoe: Changed

simulator/src/jack/listener/vm.writer.listener.ts
happytomatoe: Added assertExists function calls to satisfy compiler

simulator/src/jack/listener/vm.writer.listener.ts
happytomatoe: May I ask why it's done this way?

simulator/src/jack/listener/common.ts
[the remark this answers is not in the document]
DavidSouther: It's... complicated. TypeScript Enum syntax was an early addition to the language, that would not be allowed today. It adds behavior that isn't present in JavaScript, and achieves it by writing weird surjective object literals. There are a couple places where that's a good thing, but generally using string literal unions as enums (which don't use the `enum` keyword) are a little bit cleaner.

More importantly, the preferred way in this project.

https://www.typescriptcourse.com/string-literal-unions-over-enums for more reading.

happytomatoe:
> @happytomatoe I hope you're able to keep making progress on this, but I wanted to let you know that I'm unavailable for the coming week. I'll be back October 20th to review any changes.

Have a nice holiday)

simulator/src/jack/listener/vm.writer.listener.ts
DavidSouther: I'm not sure I understand what you're asking for an explanation on?

Why does VmInstruction exist? As an interface describing the structures in a valid Jack VM program.

Why would I suggest using the VmInstruction[] type over string? Because you retain more information for longer in the lifetime of the program. By going directly to string, you immediately discard significant amounts of information that the program already computed.

simulator/src/jack/error.ts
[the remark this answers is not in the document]
happytomatoe: Let's use Diagnostics interface.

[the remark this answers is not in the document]
DavidSouther: You probably don't want the entirety of [`Diagnostic`](https://github.com/microsoft/TypeScript/blob/a53c37d59aa0c20f566dec7e5498f05afe45dc6b/src/server/protocol.ts#L2570), just the subset that I included in the other PR.

"That's hard to read" is subjective - what do you find hard to read about it? Happy to iterate on it, but I need to know what you're pushing back on.

simulator/src/jack/listener/vm.writer.listener.ts
happytomatoe: Why we do we use array for that?

simulator/src/jack/error.ts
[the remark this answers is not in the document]
DavidSouther: (I just noticed that I somehow missed running Prettier on that PR; does it look better with a little formatting applied?)

simulator/src/jack/listener/vm.writer.listener.ts
DavidSouther: Ah, because it's the easiest "next step" from string, so it also keeps the most similarity with your current code. 

Instead of `this.result += "    call Memory.alloc 1 \n";`,  you can have `this.result.push({op: "call", name: "Memory.alloc", nArgs: 1});`

components/src/stores/compiler.store.ts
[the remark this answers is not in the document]
happytomatoe: Let's agree that we can create another PR and standardize this across whole codebase

simulator/src/jack/error.ts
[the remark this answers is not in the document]
happytomatoe: Looks better now. I guess we won't get many compiler errors typically.

simulator/src/jack/listener/common.ts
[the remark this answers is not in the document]
happytomatoe: Luckily VS code lang server has a pattern for this
```
export declare namespace InsertTextFormat {
    const PlainText: 1;
    const Snippet: 2;
}
export type InsertTextFormat = 1 | 2;
```

happytomatoe:
Small update - I was fed up working with official antlr4 typescript target. Migrated to  [antlr4ng](https://github.com/mike-lischke/antlr4ng)

simulator/src/jack/listener/vm.writer.listener.ts
happytomatoe: This is a separate change.
Correct if I am wrong - This [function](https://github.com/nand2tetris/web-ide/blob/634686eca2bbb6c0f795bf72f1e2f440647f02e5/simulator/src/jack/compiler.ts#L63C1-L65C47) has return type `Record<string, string | CompilationError> `. Current compiler creates an array of strings and then joins the lines inside this [function](https://github.com/nand2tetris/web-ide/blob/634686eca2bbb6c0f795bf72f1e2f440647f02e5/simulator/src/jack/compiler.ts#L215).

DavidSouther:
> Small update - I was fed up working with official antlr4 typescript target. Migrated to [antlr4ng](https://github.com/mike-lischke/antlr4ng)

Have you looked at Treesitter at all? (question, not a suggestion, but I hear great things about it and might be where I'd start if I was going from scratch today).

components/src/stores/compiler.store.ts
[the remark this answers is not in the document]
DavidSouther: Ok. Please do your best in the meantime to keep this as tidy as you can.

simulator/src/jack/symbol.ts
```diff
@@ -11,6 +11,7 @@ export interface SubroutineInfo {
   type: SubroutineType;
   localVarsCount?: number;
 }
+
 export type GlobalSymbolTable = Record<string, GenericSymbol>;
```
DavidSouther, line 15: I see in several places this is used specifically as a map, rather than an ad-hoc object. Should this be `Map<string, GenericSymbol>`? It'll enforce type checking the possibly null `.get` response, but will also make it easier to check `.size` instead of `Object.keys(...)`

(Edit: this is the current interface with the `compile`, so that would also need to change. Agreed to do that leter.)

simulator/package.json
```diff
@@ -22,13 +22,14 @@
     "@nand2tetris/projects": "^1.0.0",
     "@nand2tetris/runner": "^1.0.0",
     "@types/node": "^20.14.2",
-    "antlr4": "^4.13.2",
+    "antlr4ng": "^3.0.7",
```
DavidSouther, line 25: Does this change need a change in the README?

simulator/src/jack/anltr.compiler.ts
```diff
@@ -0,0 +1,165 @@
...
+): Record<string, string | CompilationError> {
+  if (files.type == "LexerOrParserError") {
+    throw new Error("Expected tree but got a lexer or parser error");
+  }
+  const result: Record<string, string | CompilationError> = {};
+  for (const name of Object.keys(files)) {
+    result[name] = "";
+  }
+  const trees: Record<string, ProgramContext> = {};
+  const errors: Record<string, CompilationError> = {};
```
DavidSouther, line 55: For most of these - are these Objects or Maps? (It's OK if they're records, but please leave a comment if that's a specific design choice, and why Record is preferred to Map)

simulator/src/jack/anltr.compiler.ts
```diff
@@ -0,0 +1,165 @@
...
+    filename?: string,
+  ): string | JackCompilerError[] {
+    const treeOrErrors = this.validate(tree, filename);
+    if (Array.isArray(treeOrErrors)) {
+      const errors = treeOrErrors as JackCompilerError[];
+      console.log("Errors in validator " + JSON.stringify(errors));
+      return errors;
+    }
+    const validateTree = treeOrErrors as ProgramContext;
+    const vmWriter = new VMWriter(
```
DavidSouther, line 138: Ahh, I added a comment earlier about Record vs Map, but VMWriter wants a Record. Hmm, maybe that should have been a Map too. Up to you to change to using Maps, or acknowledging the comment and moving on.

simulator/src/jack/antlr.compiler.test.ts
DavidSouther, line 1: This could be a fancy "golden file" test, where each jack/vm pair are their own files, which are loaded and discovered at runtime... but that's a lot of work and this is working for now. If anyone sees this review comment in the future, you might go for it then.

simulator/src/jack/listener/vm.writer.listener.ts
DavidSouther: Well... I should have made this comment there! Fine leaving the interface and we can come back to it later.

DavidSouther (changes requested):
Looking great, these are mostly optional thoughts at this point, up to you to incorporate them now or hold off a bit.

simulator/package.json
happytomatoe: I've already changed the readme. Both of this libraries implement the same api/interfaces

happytomatoe: This library already has a ton of comments in typescript sources so I don't know if makes sense to add anything else in terms of documentation

simulator/src/jack/anltr.compiler.ts
happytomatoe: These are maps. I will change it

happytomatoe:
> > Small  update  -  I  was  fed  up  working  with  official  antlr4  typescript  target.  Migrated  to  [antlr4ng](https://github.com/mike-lischke/antlr4ng)
> 
> Have you looked at Treesitter at all? (question, not a suggestion, but I hear great things about it and might be where I'd start if I was going from scratch today).

What do you hear specifically? I don't see typescript API on their page. So it's the same pain as using official antlr javascript library. And it's also a pure C library with language bindings. I don't know if this has it's own issues. 
I've seen this tool when I was building prettier plugin for Jack as it can build syntax tree(Tree with whitespaces and comments.)

If you would build this from scratch, based on my experience, the most go to options are 
- Antlr - because it's old and mature. There are commercial companies that use  this tool - [https://strumenta.com/](https://strumenta.com/). There are a lot documentation/resources that can help with specific problem. On the main website looks like Python used this library before switching to own implementation.
- Chevrotain - I've seen java prettier plugin and probably of others prettier plugins that use chevrotain. If comparing with previous option it probably is faster (according to their benchmark). Though we probably don't need a faster lexer/parser. I don't know how mature is this library. I've seen a PR where they are implementing a feature from ANTLR. I also don't know if there are commercial companies that use it.

simulator/src/jack/anltr.compiler.ts
happytomatoe: @DavidSouther  Why to choose Map in these cases?

happytomatoe: After doing a small research on this topic I don't see any substantial difference other that a little bit better api.

simulator/src/jack/antlr.compiler.test.ts
happytomatoe: For whomever reads this comment - this is partially done if you look inside jack/resources/test folder. I didn't want to do this for small code snippets because I am lazy and don't want to be jumping across files in the editor.

@DavidSouther Let me know if you want me to change the path for jack/resources/test.

components/src/stores/compiler.store.ts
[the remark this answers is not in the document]
happytomatoe: Let me know if you want me to change something. Otherwise I will leave it as it is

happytomatoe:
Testing is done from my subjective point of view.

simulator/src/jack/antlr.compiler.test.ts
DavidSouther: All good.

simulator/src/jack/symbol.ts
```diff
@@ -0,0 +1,131 @@
...
+
+type VariableType = string;
+export enum ScopeType {
+  Static,
+  This,
+  Argument,
+  Local,
+}
+
+const scopeTypeToStringMap: Record<ScopeType, string> = {
```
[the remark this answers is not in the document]
DavidSouther: I think `ScopeType` could have been `export type ScopeType = 'static'|'this'|'argument'|'local'';`

Then we wouldn't have needed scopeTypeToStringMap at all. Not a big deal wither way.

DavidSouther (approved):

PR #484, opened 2024-10-04
Run lint and format before every commit using husky

happytomatoe:
It's also possible to run this before git push. Don't think that adding test run is a good idea because of the tests execution time


DavidSouther:
It is a developer's responsibility to configure their environment to apply formatting, fixes, etc. The CI pipeline is to ensure it has been done, but the project does not mandate how that is achieved.

Issue #498, opened 2024-11-19
[bug]: VM emulator doesn't allow empty strings.


dorianignee:
### Tool

VM Emulator

### Interface

Website (https://nand2tetris.github.io/web-ide)

### Contact Details

dorianignee@hotmail.com

### What happened?

The VM emulator throws ERR14 when working with an empty string.

Reproduction steps:
1. paste the following code into a Main.jack file in the jack compiler:
```jack
class Main {
    function void main() {
        var String empty;
        let empty = "";
        return;
    }
}
```
2. Click "Compile", then click "Run"
3. When the VM emulator opened up, press the [>>] button to start the simulation

Workaround:
Since strings start empty and only contain characters, when they are appended using the String.appendChar(c) function, you can create an empty string with `let empty = String.new(1);`.

### Do you want to try to fix this bug?

- [X] I want to try to add this feature!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #499, opened 2024-11-19
fix: #498 allow Zero-sized strings in VM.

dorianignee:
The vm call `String.new` now only checks if `size` is negative, so that empty strings work as expected.


DavidSouther:
Who are we to say that a Jack programmer is allowed to create a string that is always empty? (The argument to String.new is the _max_ size, not the initial size. The initial size is always zero. So calling `String.new(0)` in Jack (or `push 0\ncall String.new 1` in VM) gives back a string that can never have any characters in it.)

PR #522, opened 2024-12-04
Update sorting for chips projects

Awkewainze:
https://github.com/nand2tetris/web-ide/commit/eb5f60496a2490c82e1ac91d4de5943038d42282 broke recommended order for chips projects after project 01 (e.g. 2, 3, 5) and instead alphabetically sorts them.

See 
image

Correct/Fixed output
image


DavidSouther (approved):
Thank you for catching this, and providing a quick update!

PR #569, opened 2025-03-19
Added a new clear grid button functionality to enhance user experience when …

Mereck600:
…drawing new bitmaps. When trying to use the bitmap for one of the projects, I found an issue where the bitmap was not displayed on the screen upon first opening the bitmap editor. This is likely due to something in the initialization, however, adding a console log (initially used for debugging) in the Init function solved this issue. I think it just bought enough time. That is not an optimal solution however, it works for now.

### DIFF
--- a/web/public/bitmap_editor.html
+++ b/web/public/bitmap_editor.html
@@ -75,6 +75,9 @@
 
       function Init() {
         grid = InitGrid();
+         //Added this console log as a simple fix to ensure the grid had time to populate, on screen the grid would occasionally not populate properly unless I cleared local storage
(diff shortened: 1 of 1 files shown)


ErikUmble:
I also noticed this problem and will look into fixing it with perhaps a more robust approach.

ErikUmble:
Yes, you were right about the issue being in the initialization. I think the problem was that the startup code assumes grid exists in localStorage while it might not actually. I decided to submit a different pull request (#579) because there were other features (such as importing png and bmp files) that should really be brought over to this platform. That pull request should resolve the loading issue.

DavidSouther:
Closing in favor of #579

PR #579, opened 2025-05-23
Update bitmap editor to v2.9

ErikUmble:
I noticed an issue with the bitmap editor not loading the canvas correctly on startup. While fixing the issue, I realized this version is behind the latest version [here](https://github.com/ErikUmble/JackBitmapEditor) and wanted to update it to the current.
New features brought in:
* ability to load .bmp and .png files into bitmaps
* improved resizing capabilities (preserves the drawing when possible during a resize)

Considering most users will be accessing the bitmap editor through this Web IDE, I will try to submit pull requests more quickly in the future, if we get other updates. I'm also open to considering the possibility of archiving the separate repository and redirecting contributors/maintainers here for future bitmap editor updates if that would be preferable.


ErikUmble:
Please note: it was [brought to my attention](https://github.com/ErikUmble/JackBitmapEditor/issues/7) that there were still loading bugs in v2.8. I've reviewed the initialization code and tested in additional browsers, and think the issue is actually fixed now in v2.9. I added a commit to this pull request to make those changes to sync this with v2.9.

Issue #428, opened 2024-08-15
[feature]: Success turns green


hannah-lau:
### Tool

General

### Interface

Website (https://nand2tetris.github.io/web-ide)

### What feature are you proposing?

Is anyone looking for a quick win? ;) I think it would be really nice if the 'simulation success' bar turned green when you succeed. Conversely, the 'simulation error' bar would turn green. Obviously isn't required for the IDE to work well, but it would add just that extra little bit of dopamine when something works correctly :)
<img width="1429" alt="Screenshot 2024-08-14 at 9 09 45 PM" src="https://github.com/user-attachments/assets/e481edf6-41dc-4f55-85b5-a4121a193d5e">

### Do you want to try to add this feature?

- [ ] I want to try to add this feature!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

DavidSouther:
@hannah-lau this sounds like a really interesting feature to add! 

The details in the status bar run through the `setStatus` function defined in the [BaseContext](https://github.com/nand2tetris/web-ide/blob/eb30f7148b0950e266a7bb9560b243becdc98f32/components/src/stores/base.context.ts#L31). This function currently takes a single parameter, a string. You could modify it to take a `string|???`, which captures the severity of the message, perhaps `SUCCESS`, `WARNING`, `ERROR`, and `INFO`? Then use the CSS variables from the theme for these background colors. I would love to see the PR!

PR #581, opened 2025-06-09
Add a status line background color

kochaika:
Adding this feature: https://github.com/nand2tetris/web-ide/issues/428

- Introduces 4 levels of status messages: SUCCESS, WARNING, ERROR, and INFO
- The background color for status line is changed to: green, yellow, red respectively and is not changed for INFO messages
- The default level for `setStatus` is considered INFO, backward compatibility is kept
- Both color schemes are supported

Screenshot 2025-06-09 at 14 35 33
Screenshot 2025-06-09 at 14 35 49

### DIFF
--- a/web/src/shell/statusline.scss
+++ b/web/src/shell/statusline.scss
@@ -1,0 +1,33 @@
+:root[data-theme="light"] {
+  --status-info-bg-color: inherit;
+  --status-success-bg-color: #aceebb;
+  --status-warning-bg-color: #f4e5b1;
+  --status-error-bg-color: #ffc1c0;
+}
+
+:root[data-theme="dark"] {
+  --status-info-bg-color: inherit;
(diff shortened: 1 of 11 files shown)


DavidSouther:
Fixes #428

web/src/shell/statusline.scss
```diff
@@ -0,0 +1,33 @@
+:root[data-theme="light"] {
+  --status-info-bg-color: inherit;
+  --status-success-bg-color: #aceebb;
```
DavidSouther, line 3: Where did these colors come from? It would probably be better to use the Pico color variables - https://github.com/picocss/pico/blob/v1-dev/scss/themes/default/_colors.scss

(I wish Pico was a bit more opinionated about status line colors, but alas)

DavidSouther (changes requested):

Issue #565, opened 2025-03-02
[bug]: Project 4 - Web IDE - CPU Emulator- Running Mult.tst and Fill.tst deletes user ROM input


FlorianMiedtke:
### Tool

CPU Emulator

### Interface

Website (https://nand2tetris.github.io/web-ide)

### What happened?

When I want to run the test files for project 4, they delete the contents of the ROM. This seems to happen because of the line 10 of Mult.tst

https://github.com/nand2tetris/web-ide/blob/cd2963f4ad14c1f8ce99ae0f11d5a749cc28bc0c/projects/src/project_04/01_mult.ts#L21

This seems to load the default Mult.asm file of the Web IDE file system, which is always the empty default. An easy fix would be to just comment out this line, but I think this would ignore the underlying problem. 

If you look at the hardware simulator, the code written there is saved to the Web IDEs file system. So when I reload the page, my changes are still there. In the CPU Emulator, the ROM is also cleared, when I reload the page. So the actual problem seems to be, that the user input in the ROM is not saved to the Web IDEs file system.

amama22:
having the same problem. how did u manage to test it?

FlorianMiedtke:
> having the same problem. how did u manage to test it?

I just executed the defined test steps manually.

jimmyvh:
Just ran into this issue myself, and found this bug after googling.
As a workaround, executing the fist step first, then filling out the asm code works, but it's an annoying workaround.
The real issue is that the asm code is not being saved. 
Test failing and you need to make fixes? Fill in everything from scratch. Want to to let it sink into your brain for a while and come back later? Start from scratch. Could use an external text editor to save inbetween ofcourse, but you still have to manually type over everything between each iteration.
If the file actually gets saved like it does for the hardware simulator, everything would work as intended, the *load Mult.asm* line included.

gacarrillor:
Same issue here.

Duplicate of #551 and #543

amama22:
> Same issue here.
> 
> Duplicate of [#551](https://github.com/nand2tetris/web-ide/issues/551) and [#543](https://github.com/nand2tetris/web-ide/issues/543)

Hey, if this helps, I realized after, in the assembler u have to load the asm file lol and fill that, and then when u load that file in the cpu emulator it will correctly open that same file.

chids04:
can confirm this issue is still occuring

Issue #551, opened 2025-01-05
[bug]: CPU Emulator: Cannot update Mult.asm or test local file with Mult.tst


joe-tripodi:
### Tool

CPU Emulator

### Interface

Website (https://nand2tetris.github.io/web-ide)

### What happened?

I want to run the Mult.tst against my local file.
I load the local file
I load the test
The ROM gets wiped and I cannot test my local file against the Mult.tst

I attempt to edit the ROM from the online editor instead
I type in a command but when I tab or press enter, it gets wiped.

### Do you want to try to fix this bug?

- [ ] I want to try to add this feature!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

direvus:
I'm also experiencing this issue.

I suspect that the problem might be this "load Mult.asm" instruction in the test script https://github.com/nand2tetris/web-ide/blob/cd2963f4ad14c1f8ce99ae0f11d5a749cc28bc0c/projects/src/project_04/01_mult.ts#L21

If it's trying to load Mult.asm from the project files, that would explain why the program ROM gets wiped -- Mult.asm in the project files is empty.

I think this bug means that test scripts are basically unusable in the CPU Emulator mode of the Web IDE.

Suggest that the "load" instruction of the test script in the IDE should have the effect of "read the contents of the ROM panel of the IDE, and load it into the program ROM of the emulated machine". It definitely should not ever replace or modify the user's input in the ROM panel.

Issue #543, opened 2024-12-25
[bug]: Cannot write to Mult.asm file


petrovj12:
### Tool

CPU Emulator

### Interface

Website (https://nand2tetris.github.io/web-ide)

### What happened?

I cannot write to the file Mult.asm, neither by editing it directly in the CPU Emulator nor by transferring from the Assembler. When the Mult.tst script tries to load the file Mult.asm it is loaded as only zero instructions.

### Do you want to try to fix this bug?

- [ ] I want to try to add this feature!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

pbrw:
I faced the same issue. I wrote assembler code in the Web UI. Then I run a `Mult.tst` test but the line `load Mult.asm` just cleared my ROM and I can't find a way to get my solution back.

maueroats:
Clarification: in Project 4, the testing code begins with `load Mult.asm`, but when testing this code, the user will be uploading their `Mult.asm` file written somewhere else and either uploaded or copy-and-pasted into the left side code pane.  

Under these circumstances, the code that users want to test is not "in the file" Mult.asm. 

I see that the code *does* revert to some "saved" version.

Somehow, though, my "saved" version is something I myself wrote. I don't know what I did to make that happen. Perhaps the behavior changed recently? I see the ability to edit the test scripts seems to have disappeared, or there would be an easy fix - "comment out the load statement".

maueroats:
Please rename bug to be more informative: project 4 testing code re-loads asm files without including user changes

t1mp4:
> Clarification: in Project 4, the testing code begins with `load Mult.asm`, but when testing this code, the user will be uploading their `Mult.asm` file written somewhere else and either uploaded or copy-and-pasted into the left side code pane.
> 
> Under these circumstances, the code that users want to test is not "in the file" Mult.asm.
> 
> I observed that the assembly code _does_ revert to some "saved" version when the test script is run. In a "clean" browser, this file is empty / full of zeroes.
> 
> In one of my browsers, my "saved" version is something I myself wrote. I don't know what I did to make that happen. Perhaps the behavior changed recently?
> 
> I see the ability to edit the test scripts seems to have disappeared, or there would be an easy fix - "comment out the load statement".

Yeah, what's the reasoning behind disallowing the user to edit test scripts or upload their own test scripts?

t1mp4:
[...]

PR #591, opened 2025-08-13
Allow users to upload local .tst and .cmp files in CPU Emulator

HaorongX:
As described in issues #543 #565 #551 and #543, using the built-in Mult.tst file always overwrite users' programs in RAM, which makes CPU Emulator completely unusable in project 4. Therefore, I changed loadTest such that:

- It allows users to upload local .cmp and .tst file

Benefits:

- Users can use their .tst file without the problematic `load Mult.asm`
- The current fs-based test loading is maintained

However, it requires users to load .cmp and .tst separately, and `compare-to Mult.cmp` must be removed simultaneously to use local .cmp file.

### DIFF
--- a/web/src/shell/test_panel.tsx
+++ b/web/src/shell/test_panel.tsx
@@ -28,6 +28,7 @@ import { AppContext } from "../App.context";
 import { Editor } from "./editor";
 import { Panel } from "./panel";
 import { Tab, TabList } from "./tabs";
+import { isPath } from "src/shell/file_select";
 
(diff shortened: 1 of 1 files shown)


web/src/shell/test_panel.tsx
```diff
@@ -295,4 +313,4 @@ function decorationTypeToCss(type: DecorationType) {
     default:
       return "";
   }
-}
+}
```
DavidSouther, line 316: (nit) ensure formatting has the last line end

PR #596, opened 2025-09-17
Added manual to bitmap editor and also small changes at the bitmap html

renanbazinin:
the commit of the bitmap edtior html looks long but it mostly moving parts in the UI there....


how -U0 0c1ed467
```diff
@@ -0,0 +1,11 @@
+[33mcommit 0c1ed467e1f75956bd2840fc95e9610c56db5887[m[33m ([m[1;36mHEAD -> [m[1;32mmain[m[33m, [m[1;31morigin/main[m[33m, [m[1;31morigin/HEAD[m[33m)[m
```
DavidSouther, line 1: What is this file?

web/public/bitmap_editor.html
```diff
@@ -27,22 +27,20 @@
         baseJ: 0,
         drawHeight: 0,
         INVERT_MODE: false,
-        comments: true,
```
DavidSouther, line 30: Please make sure you're using Biome as your code formatter. You can also run `npm run fix` (I think... look in package.json for the specific scripts)

web/public/user_guide/bitmap editor.pdf
DavidSouther, line 1: No spaces in file names, and is this a duplicate of the other?

DavidSouther (changes requested):
This looks like two changes- the updated user guides, and then the bitmap editor hack/jack switcher. Please separate the later into a different PR. Let me know if you need instructions on how to do that, I'm not sure your level of Git familiarity.

PR #624, opened 2025-12-22
Switch to Vitest & bump Biome

DavidSouther:
Switch from Jest to Vitest, the primary benefit being native JSX testing.


DavidSouther:
Completed in #646

PR #627, opened 2026-01-01
Bitmap editor Refactor From HTML

renanbazinin:
Hey, 
Took @ErikUmble 's (Thank You) original HTML file for the Bitmap Editor and refactored it into React.

### DIFF
--- a/web/src/pages/bitmap.tsx
+++ b/web/src/pages/bitmap.tsx
@@ -344,8 +344,8 @@ export const BitmapEditor = () => {
     return edges;
   }, [obviousStyling, getUsedWords]);
 
-  // Event handlers
-  const handleCellMouseDown = useCallback(
+  // Drawing event handlers
+  const startDraw = useCallback(
     (i: number, j: number) => {
       // use functional update to avoid reading stale grid from closure
       setGrid((prev) => {
@@ -359,7 +359,7 @@ export const BitmapEditor = () => {
     []
   );
 
-  const handleCellMouseOver = useCallback(
+  const moveDraw = useCallback(
     (i: number, j: number) => {
       if (currentColor === null) return;
       setGrid((prev) => {
@@ -371,7 +371,7 @@ export const BitmapEditor = () => {
     [currentColor]
   );
 
-  const handleMouseUp = useCallback(() => {
+  const stopDraw = useCallback(() => {
     setCurrentColor(null);
   }, []);
 
@@ -629,7 +629,7 @@ export const BitmapEditor = () => {
   }, []);
 
   return (
-    <div className="Page BitmapPage grid" onMouseUp={handleMouseUp} onMouseLeave={handleMouseUp}>
+    <div className="Page BitmapPage grid" onMouseUp={stopDraw} onMouseLeave={stopDraw}>
       {/* Canvas Panel */}
       <Panel
         className="canvas-panel"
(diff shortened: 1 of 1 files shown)


ErikUmble:
Cool! When this is merged, I'm happy to archive the [old repo](https://github.com/ErikUmble/JackBitmapEditor) so that future development on the bitmap editor can be done here.

web/src/pages/bitmap.scss
```diff
@@ -0,0 +1,236 @@
...
+        min-width: 400px;
+    }
+
+    .controls-panel {
+        grid-area: controls;
+    }
+
+    // Canvas size header controls
+    .canvas-size-controls {
+        display: flex;
```
DavidSouther: I generally like having an explicit flex-direction because I honestly never remember that the default is probably "row"

web/src/pages/bitmap.tsx
```diff
@@ -1,10 +1,1030 @@
...
+  drawHeight: 0,
+  invertMode: false,
+  comments: true,
+};
+
+type CodeMode = "jack" | "hack";
+type MarginType = "fitToDrawing" | "rectangular" | "fullCanvas";
+
+// Utility functions for code generation
+function getWordValue(
```
DavidSouther: Could we pull these to another file and add some unit tests for them?

web/src/pages/bitmap.tsx
```diff
@@ -1,10 +1,1030 @@
...
+          return parsed;
+        }
+      }
+    } catch (e) {
+      console.error("Error loading bitmap editor grid:", e);
+    }
+    return createGrid(width, height, invertMode);
+  };
+
+  const createGrid = (width: number, height: number, invertMode: boolean): boolean[][] => {
```
DavidSouther: Could probably live with the utils, above.

web/src/pages/bitmap.tsx
```diff
@@ -1,10 +1,1030 @@
...
+    const { columns, baseJ } = getUsedWords();
+    const edges = new Set<number>();
+    for (const col of columns) {
+      edges.add(col * 16 + baseJ);
+    }
+    return edges;
+  }, [obviousStyling, getUsedWords]);
+
+  // Event handlers
+  const handleCellMouseDown = useCallback(
```
DavidSouther: I would rename these as `startDraw`, `moveDraw`, and `stopDraw`, so that later you can assign stopDraw to several events including mouseUp and mouseOut (and in the future, we can use them for touch events)

web/src/pages/bitmap.tsx
```diff
@@ -1,10 +1,1030 @@
...
+  // keep uncontrolled input refs in sync when state changes (consistent with other pages)
+  useEffect(() => {
+    if (widthInputRef.current) widthInputRef.current.value = width.toString();
+    if (heightInputRef.current) heightInputRef.current.value = height.toString();
+    if (pixelSizeInputRef.current) pixelSizeInputRef.current.value = pixelSize.toString();
+    if (marginFramesInputRef.current) marginFramesInputRef.current.value = marginSaveFrames.toString();
+  }, [width, height, pixelSize, marginSaveFrames]);
+
+  // File import handlers
+  const handleFileSelect = useCallback(
```
DavidSouther: These file and parse operations should probably be in another file, and work in conjunction with this rather than be directly inline.

web/src/pages/bitmap.scss
```diff
@@ -0,0 +1,236 @@
...
+        flex-wrap: nowrap;
+        gap: var(--spacing);
+        align-items: center;
+        padding: var(--spacing);
```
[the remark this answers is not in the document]
renanbazinin: gap/padding will add a bit of whitespace, but it’s intentionally localized to the control group :

image

Without them the controls end up looking jammed together (Like here):
image

If you prefer it tighter / more explicit, I can reduce the spacing, Hard coded or aligned :

gap: 4px
padding: 4px

Or:

gap: calc(var(--spacing) * 0.5)
padding: calc(var(--spacing) * 0.5)

PR #649, opened 2026-04-26
Editor e2e PageObject

DavidSouther:


mmatsumoto1026:
After executing `chipPage.editor.write` in the test "toggleMonaco re-enables Monaco editor for tests that need it" of editor-panel.spec.ts, the code text in the Monaco editor written by key press simulator will break as shown in below.

```
CIP nd
	IN, b;
	  outot;

		PARTS:
		   Nandaa = ot=);
		       Not(n=,out=ut);
		       }
```

The method `this.page.keyboard.type(text)` in Monaco may triggers several automated features like auto-completion.

With using `this.page.keyboard.insertText(text)`, We can avoid the individual key-event logic for most editor side-effects, allowing the entire block of text to be inserted cleanly into the Monaco.

